PR #99, opened 2014-11-25
Made many fields work in group

asknaskan:
+ Added full support for wysiwyg fields and init of QTags (group lists)
+ Added full fix for ids, classes and values if an new group row will
create (prevent first creation bug of selects, radios and checkboxes)
+ Make ShiftRow work for wysiwyg, checkboxes, selects, radios, color
pickers (untested but should work as before because there is no change: time pickers, datepickers)

#cleanRow : Handling for Divs added, Filter for Wysiwyg added, Label
handling changed, wysiwyg handling changed, id and class handling
changed
#afterRowInsert : Added handling for init wysiwyg tinymce and qtags,
handling to call tinymce or qtags based on last state or activated mode
(wp global: no wysiwyg)
#shiftRows : Added handling for wysiwyg, added handling for radios,
changed handling for selects
#initColorPickers : Added reinit handling after color changed


asknaskan:
Merged to current upstream. 

Known Bugs last in CMB2 I see:
- Deleting the last Group and create an new again will cause an javascript crash
- Deleting an group is not last will destroy the next new created group
- (there should be an way to unload preloaded tinymce editors in code for deleting things)

I will see the next couple of days.

DevinWalker:
@asknaskan nice pull request. I'm looking forward to this merge also.

Blindmikey:
Looking forward to this merge, sounds like it could fix https://github.com/WebDevStudios/CMB2/issues/148

asknaskan:
Thanks for the feedback.

Im currently thinking about how to prevent order errors by deleting groups that not at end of row. I will try the next couple of days to find a workable solution (and some time).

asknaskan:
Everything "silent merged" as needed.

PR #120, opened 2014-12-15
file and file_list types work like "Featured Image"

johnsonpaul1014:
Made file and file_list types for images work like the "Featured Image" in Wordpress: clicking on the image brings it up in the Media Manager.  I made this change because I knew the user would possibly want to edit meta information for images, and the old version that just saved the URL made it so the user would need to look up the image in the library again.  I was unable to get the image uploader to add the image to the search results like the featured image select does, but it will appear in the box on the right.

This is my first pull request on a project on GitHub, so I apologize if anything I did was incorrect.  I believe I followed all of your guidelines, but I may have missed something.  Hopefully you like the change so I can easily get updates in the future; it's a great extension.  Thanks in advance for taking a look.


jtsternberg:
Thanks for your PR. I actually tackled this the other day without realizing you had done most of the work already. I gave you props, as you submitted a PR that is close to what I ended up with, and after I found your PR, I was able to improve a couple things in mine. Thanks again!

PR #138, opened 2015-01-07
Added support for radio input swapping in repeatable fields

DevinWalker:
I noticed that my radio input in a repeatable field wasn't working properly when shifting rows up & down. After a simple change to cmb.js it appears working find. I've run grunt through and then submitted this pull req. Let me know if you have any questions!


jtsternberg:
Nice catch, thanks!

PR #145, opened 2015-01-14
Fix metabox form not being returned to caller

akshayagarwal:


jtsternberg:
Good catch. Alternatively you could (and maybe should?) use the `cmb2_get_metabox_form` function to get a metabox object. Thank you for your submission!

PR #149, opened 2015-01-17
Minor JS addition to full support radios in repeatable field groups

DevinWalker:
I realized from my previous pull request I missed another check for is :checked when testing. This pull request not accounts for radio elements in the $parent find loop.


jtsternberg:
nice catch, thanks!

PR #151, opened 2015-01-19
Accept value '0' for select, radio field

inspiraaz:
Problem: https://github.com/WebDevStudios/CMB2/issues/115

Current isempty function returns true with value '0'.

With this code we can accept both '0' and 0.
empty( $value ) && $value != '0'

Then we can properly check if value is empty.

Works with select and radio field.
Multicheck field still doesn't work if we set value to 0.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -201,9 +201,13 @@ class CMB2_Types {
 		$method = isset( $args['method'] ) ? $args['method'] : 'select_option';
 		unset( $args['method'] );
 
-		$value = $this->field->escaped_value()
-			? $this->field->escaped_value()
-			: $this->field->args( 'default' );
+	        if( cmb2_utils()->isempty( $this->field->escaped_value() ) ) {
+	            $value = $this->field->args( 'default' );
+	            $value_isempty = true;
+	        } else {
+	            $value = $this->field->escaped_value();
+	            $value_isempty = false;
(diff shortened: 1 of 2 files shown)


jtsternberg:
I think this issue has been resolved. Can you please update your copy of CMB2 and verify?

inspiraaz:
CMB 2.0.1
Wordpress 4.1

It works now, value '0' is saved.

But there is stil a problem with selected attribute applied to both value '0' and empty value.

metabox example:

```
array(
    'name' => 'My Field',
    'id'   => 'my_field',
    'type' => 'select',
    'options' => array(
        ''   => 'Select',
        '0'  => 'No',
        '1'  => 'Yes'
    )
)
```
If i select option with value '0' or value '', then the html output looks like this

```
<select>
    <option value="" selected='selected'>Select</option>
    <option value="0" selected='selected'>No</option>
    <option value="1" >Yes</option>
</select>
```

jtsternberg:
I think the simple solution would be to not use the value `0` as your option value, but use a string instead, like `'no'`.  Then in your output, if you really need the `0` value, you can check for `'no'` and translate it to `0`. 
I'll leave this issue open though as this is a bug.

tw2113:
Closing along the same lines as #557 and https://github.com/WebDevStudios/CMB2/pull/557#issuecomment-236802807

PR #162, opened 2015-01-27
Run stripslashes before saving data, since WordPress forces magic quotes

clifgriffin:
We found that when entering a single tick `'` into text fields, the resulting save adds an unnecessary escape character: `\'`

This is because of that weird legacy quirk with WordPress that forces magic quotes onto all incoming data, whether `$_GET` or `$_POST`. This happens regardless of the PHP setting so trying to use `get_magic_quotes_gpc()` in this context does not work. (See note here: http://codex.wordpress.org/Function_Reference/stripslashes_deep) 

This patch uses stripslashes_deep to nuke all of these befor storing fields in the database. We've tested and it works perfectly.

### DIFF
--- a/CMB2.php
+++ b/CMB2.php
@@ -391,7 +391,7 @@ class CMB2 {
 	 */
 	public function save_fields( $object_id = 0, $object_type = '', $data_to_save ) {
 
-		$this->data_to_save = $data_to_save;
+		$this->data_to_save = stripslashes_deep($data_to_save);
 		$object_id = $this->object_id( $object_id );
 		$object_type = $this->object_type( $object_type );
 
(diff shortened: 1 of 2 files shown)


jtsternberg:
The thing that concerns me with deep changes like this is the unintended side-effects for existing users. Can you think of any reasons why/how this would cause problems?

jtsternberg:
Also, if we implement this, it most likely should go in the CMB2_Sanitize constructor.

clifgriffin:
@jtsternberg I think this is a safe change because we're only dealing with input.  Everyone who has input that is already badly escaped will have to fix it manually anyway. 

As far as putting it in CMB2_Sanitize, I agree and have moved it thusly.

jtsternberg:
Thanks @clifgriffin!

PR #172, opened 2015-02-06
List none

phh:
New arg: option_none. For select and radio boxes. Useful to have this argument instead of adding an extra "empty" value into the options array.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -205,7 +205,20 @@ class CMB2_Types {
 			? $this->field->escaped_value()
 			: $this->field->args( 'default' );
 
-		$concatenated_items = ''; $i = 1;
+		$concatenated_items = '';
+
+		// Check for option_none
+		if( $this->field->args( 'option_none' ) ) {
+			$a = $args;
+
+			$a['value'] = 0;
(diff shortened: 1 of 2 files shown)


jtsternberg:
Can you explain why you think:
```php
array(
	'name'        => 'Test Select',
	'id'          => $prefix . 'test_select',
	'type'        => 'select',
	'option_none' => __( 'None', 'cmb' ),
	'options'     => array(
		'standard' => __( 'Option One', 'cmb' ),
		'custom'   => __( 'Option Two', 'cmb' ),
		'none'     => __( 'Option Three', 'cmb' ),
	),
),
```
is better than:
```php
array(
	'name'    => 'Test Select',
	'id'      => $prefix . 'test_select',
	'type'    => 'select',
	'options' => array(
		'' => __( 'None', 'cmb' ),
		'standard' => __( 'Option One', 'cmb' ),
		'custom'   => __( 'Option Two', 'cmb' ),
		'none'     => __( 'Option Three', 'cmb' ),
	),
),
```
?

phh:
Was just a way to add a callback to another function pulling posts out without having to add your own empty value.

PR #173, opened 2015-02-06
use attachment_url_to_postid instead of custom url to id function

phh:
Sorry - trying again :)

In WordPress 4.0.0 attachment_url_to_postid was added.

I could remove the `function_exists( 'attachment_url_to_postid' )` check, but that would require the project to bump its minimum WP version.

### DIFF
--- a/includes/CMB2_Utils.php
+++ b/includes/CMB2_Utils.php
@@ -20,6 +20,11 @@ class CMB2_Utils {
 	 * @return mixed            Attachment ID or false
 	 */
 	public function image_id_from_url( $img_url ) {
+		// Since WP 4.0.0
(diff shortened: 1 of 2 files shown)


jtsternberg:
If you fix the parse error from the missing closing parenthesis in the function_exists conditional, i'll merge this in. :)

phh:
Updated. But seems like its still failing a test?

jtsternberg:
Yes, the test that is failing:
> 1) CMB2_Utils_Test::test_image_id_from_url
> Failed asserting that 148 matches expected null.

The test only fails on WP installations 4.0.0 or later (i.e. versions which have the `attachment_url_to_postid` function). This means the functions probably doesn't do what you think it does, or that it's not a direct replacement.

PR #190, opened 2015-02-13
Better box instantiation

jtsternberg:


jtsternberg:
Per @JayWood, need some sort of hook to be able to remove boxes/fields.

PR #198, opened 2015-02-16
Add autoload to the composer.json

mchiron:
This remove the need to do this:

```
if ( file_exists(  __DIR__ . '/cmb2/init.php' ) ) {
  require_once  __DIR__ . '/cmb2/init.php';
} elseif ( file_exists(  __DIR__ . '/CMB2/init.php' ) ) {
  require_once  __DIR__ . '/CMB2/init.php';
}
``` 

Using the autoload from composer is the right thing to do.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,5 +25,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.*"
+    },
+    "autoload": {
+        "files": ["init.php"]
     }
 }


mAAdhaTTah:
This was actually removed previously. See this convo for why this doesn't work as a solution:

https://github.com/WebDevStudios/CMB2/commit/e4d03897689a16d0231c2a344fc15b0439e2155f

jtsternberg:
@mchiron What @mAAdhaTTah said. Thank you for submitting though.

PR #214, opened 2015-02-25
Added new "text_number" field and minor style

DevinWalker:
I had a need for a "number" type input and noticed CMB2 doesn't currently support this field type.


jtsternberg:
Hmm, I'm not convinced this is necessary, since you can just add a `'attributes' => array( 'type' => 'number' )` to a `'text_medium'` field registration. Thoughts?

DevinWalker:
I see what you are saying. I guess it's mainly a convenience play. You could do some validation on this field type to ensure that it's in fact an number. Also, you have text_medium / text_small which are pretty much just class changes, so the same argument could be made for those.

jtsternberg:
TOTALLY agree about small/medium, but they are left from legacy. I'd probably not add them at this point. :) At this point, I don't really want to add more field types (without MAJOR consideration and justification), as it is more that core needs to support. I think additional field types are a great thing for custom extensions and/or additions to the [snippet library](https://github.com/WebDevStudios/CMB2-Snippet-Library). I hope my perspective makes sense.

DevinWalker:
That's cool. No problem. I get what you're saying.

PR #216, opened 2015-02-26
Add PHP7 support with allowed failures when issues come up

tw2113:


jtsternberg:
Hmm.. based on [the failures](https://travis-ci.org/WebDevStudios/CMB2/builds/52213544) (0 tests are run, because they fail instantly), it appears that we gain nothing from testing in PHP 7 until WP core updates its dependency on mysql_connect.

PR #219, opened 2015-02-26
Add group_title_callback option to customize the group title

Ramblurr:
As mentioned in the [WP.org thread](https://wordpress.org/support/topic/customize-group_title-with-callback?replies=2), it is nice to customize the title of an options group to something other than a static field with a number.

This change adds an optional callback that formats the name based off the fields.
 
kiri-mockup2

```
# add the callback parameter
'group_title' => 'New Widget',
'group_title_callback' => array($this, 'format_group_title'),
```

The title formatting function
```public function format_group_title($fields) {
        return $fields['time'] . ' ' . $fields['name'];
}
```

It's not completely ideal, as when you add an item, before clicking save, the JS formats it with the default title.

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -770,8 +770,20 @@ class CMB2_Field {
 	 * @return string        Updated value
 	 */
 	public function replace_hash( $value ) {
-		// Replace hash with 1 based count
-		return str_ireplace( '{#}', ( $this->count() + 1 ), $value );
+		if ( ! is_null( $this->escaped_value ) ) {
+		    // Replace hash with 1 based count
+		    return str_ireplace( '{#}', ( $this->count() + 1 ), $value );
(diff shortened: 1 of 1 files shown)


jtsternberg:
Sorry, this is definitely not the right way to handle this problem. Your best solution is to use Javascript, since the whole repeat-group feature is heavily-dependent on JS, and like you said, your edits will not work on the repeated groups anyway.

DevinWalker:
I'd love to see this implemented +1

jtsternberg:
A good proof-of-concept for this is [in the snippet library](https://github.com/WebDevStudios/CMB2-Snippet-Library/blob/master/javascript/dynamically-change-group-field-title-from-subfield.php), but I would like to eventually get this in core.

jtsternberg:
Related #683

PR #220, opened 2015-02-26
Add HTML class option to group fields

Ramblurr:
I did this so I could force some group options to be closed automatically, by adding the `closed` class attribute.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -287,7 +287,7 @@ class CMB2 {
 	public function render_group_row( $field_group, $remove_disabled ) {
 
 		echo '
-		<div class="postbox cmb-row cmb-repeatable-grouping" data-iterator="', $field_group->count(), '">
+		<div class="postbox cmb-row cmb-repeatable-grouping ',esc_attr($field_group->options('html_class')),'" data-iterator="', $field_group->count(), '">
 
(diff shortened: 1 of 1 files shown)


Ramblurr:
Woops, didn't see you're running unit tests. I'll work one up and add the commit.

jtsternberg:
@Ramblurr can you give me a broader/more fleshed out explanation of what you're trying to accomplish. Preferably w/ code that interacts w/ this?

Ramblurr:
@jtsternberg Absolutely. 


The goal was to automatically render the group boxes closed. Since adding the `closed` class is sufficient to due this, I added added this attribute in a generic fashion.

I can see that adding the ability to set any 'ol class isn't really necessary, perhaps it could just be a boolean option that adds the closed class conditionally?

The overriding goal between this and #219 was to produce a compact series of group fields, where (1) existing fields default to the closed state and (2) reading the group_title gives sufficient information about the group's contents. 

Like this:

sample

Here's a snippet where I use the option. It includes the group_title callback in #219, which I understand you won't merge, no worries.

```php
        $group_field = array (
            'type' => 'group',
            'options' => array(
                'html_class' => 'closed',
                'group_title' => 'New Program',
                'group_title_callback' => array($this, 'format_group_title'),
....
```

jtsternberg:
This is already implemented with the `'classes'` and `'classes_cb'` field parameters.

PR #236, opened 2015-03-07
Update CMB2_Types.php

hasinhayder:
CMB2 has a strange bug that creates an extra "select" dropdown field when the "repeatable" argument is set to true for a "select" field and it has more than 1 values. In simple, if a select field has been repeated N times where N>1, this bug always renders an extra select dropdown at the end. 

This bug is only related to "select" fields, specially when their "repeatable" is set to true. Doesn't affect other fields.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -310,6 +310,9 @@ class CMB2_Types {
 			$meta_value = $default;
 		}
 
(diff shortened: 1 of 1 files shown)


jtsternberg:
Hey @hasinhayder, as per the [contributing guidelines](https://github.com/WebDevStudios/CMB2/blob/master/CONTRIBUTING.md), can you please checkout trunk (the development branch) and make your changes and submit your PR against that branch? Thank you kindly.

PR #241, opened 2015-03-10
Fix how strip_slashes is being used

dustyf:
Patch fixes how strip_slashes_deep is being used and actually uses it.  It was added to $this->value variable in the CMB2_Sanitization class, but the methods within the class were still being passed the un-slash-stripped $value as well.  Now they will be using the value passed to the class which has been slash-stripped.

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -402,7 +402,7 @@ class CMB2_Field {
 
 		$clean = new CMB2_Sanitize( $this, $meta_value );
 		// Validation via 'CMB2_Sanitize' (with fallback filter)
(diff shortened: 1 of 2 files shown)


jtsternberg:
Thanks @dustyf, this is great, and thank you for catching it.
[I merged this](https://github.com/WebDevStudios/CMB2/commits/trunk). It was simple to merge to trunk because the edits were almost all in one file, but for future reference, please submit the pull requests to trunk, the in-development branch.

jtsternberg:
This update is a continuation of #162 which added the `strip_slashes`, but we (incorrectly) were't doing anything with that modified value.

PR #242, opened 2015-03-10
Fix for the bug that caused rendering an extra select dropdown for select fields when repeatable is set to true

hasinhayder:
CMB2 has a strange bug that creates an extra "select" dropdown field when the "repeatable" argument is set to true for a "select" field and it has more than 1 values. In simple, if a select field has been repeated N times where N>1, this bug always renders an extra select dropdown at the end.

This bug is only related to "select" fields, specially when their "repeatable" is set to true. Doesn't affect other fields.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -310,6 +310,9 @@ class CMB2_Types {
 			$meta_value = $default;
 		}
 
+		// fix the empty select dropdown bug when the repeatable is set to 'true' for a 'select' type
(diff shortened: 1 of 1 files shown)


tw2113:
@hasinhayder I realize it's been awhile, but can you confirm whether this is still an issue in current versions of CMB2?

bradp:
Also, please update this PR to match the [WordPress Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/).

tw2113:
As is, I can't recreate this, based on what information we have available. Until someone can provide steps to successfully recreate, I'm going to close.

PR #262, opened 2015-03-23
Add option to store values in individual rows instead of serialized array

### LINKED ISSUE #183: Multicheckbox - save into multiple meta instead as array

JohnyGoerend:
In previous versions, multicheck fields were stored using multiple key/value entries. This was recently changed (maybe due to a bug) so now those fields got saved as serialized arrays. Unfortunately, that change made it impossible to use the "meta_key" and "meta_value" arguments with the get_posts function. 
This patch leaves the possbility to the user with an additional "multiple" option, to store the values as individual rows again.
Fixes #183

(Re-done to correct #225)

### DIFF
--- a/example-functions.php
+++ b/example-functions.php
@@ -301,6 +301,7 @@ function yourprefix_register_demo_metabox() {
 		'desc'    => __( 'field description (optional)', 'cmb2' ),
 		'id'      => $prefix . 'multicheckbox',
 		'type'    => 'multicheck',
+		//'multiple' => true, // Store values in individual rows
 		'options' => array(
 			'check1' => __( 'Check One', 'cmb2' ),
 			'check2' => __( 'Check Two', 'cmb2' ),
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -448,17 +448,16 @@ class CMB2_Field {
 
 		$new_value = $this->sanitization_cb( $meta_value );
 		$old       = $this->get_data();
-		// $name      = $this->id();
-		// if ( $this->args( 'multiple' ) && ! $this->args( 'repeatable' ) && ! $this->group ) {
-		// 	$this->remove_data();
-		// 	if ( ! empty( $new_value ) ) {
-		// 		foreach ( $new_value as $add_new ) {
-		// 			$this->updated[] = $name;
-		// 			$this->update_data( $add_new, $name, false );
-		// 		}
-		// 	}
-		// } else
-		if ( ! cmb2_utils()->isempty( $new_value ) && $new_value !== $old ) {
+		$name      = $this->id();
+		if ( $this->args( 'multiple' ) && ! $this->args( 'repeatable' ) && ! $this->group ) {
+			$this->remove_data();
+			if ( ! empty( $new_value ) ) {
+				foreach ( $new_value as $add_new ) {
+					$this->updated[] = $name;
+					$this->update_data( $add_new, false );
+				}
+			}
+		} elseif ( ! cmb2_utils()->isempty( $new_value ) && $new_value !== $old ) {
 			return $this->update_data( $new_value );
 		} elseif ( cmb2_utils()->isempty( $new_value ) ) {
 			return $this->remove_data();


mweimerskirch:
:+1:

natebeaty:
Just dug around trying to figure out why this wasn't working after upgrading from CMB. +1 for merging this from someone that uses this functionality on a few sites.

example-functions.php
```diff
@@ -299,8 +299,9 @@ function yourprefix_register_demo_metabox() {
 	$cmb_demo->add_field( array(
 		'name'    => __( 'Test Multi Checkbox', 'cmb2' ),
 		'desc'    => __( 'field description (optional)', 'cmb2' ),
-		'id'      => $prefix . 'multicheckbox',
+		'id'      => $prefix . 'multicheckbox2',
```
jtsternberg: No need to rename this

example-functions.php
```diff
@@ -299,8 +299,9 @@ function yourprefix_register_demo_metabox() {
 	$cmb_demo->add_field( array(
 		'name'    => __( 'Test Multi Checkbox', 'cmb2' ),
 		'desc'    => __( 'field description (optional)', 'cmb2' ),
-		'id'      => $prefix . 'multicheckbox',
+		'id'      => $prefix . 'multicheckbox2',
 		'type'    => 'multicheck',
+		'multiple' => true, // Store values in individual rows
```
jtsternberg: Should comment this out by default since that is not the default behavior and is how optional arguments are noted in this file.

jtsternberg:
@JohnyGoerend @mweimerskirch @natebeaty If you all test this patch and confirm it works with and without the multiple option, I'll be able to get this merged much quicker.

natebeaty:
I can confirm it as working for me. 

I've applied the patch and when I add `'multiple' => true` it stores the array of meta values as separate entries properly in the database, and when I comment it out, it goes back to the single entry method (e.g. `a:2:{i:0;s:3:"378";i:1;s:2:"25";}`).

mweimerskirch:
Works perfectly for me as well, just as it used to with CMB1.

JohnyGoerend:
@jtsternberg Fixed.

I've already used it on a website which is online since last week.

natebeaty:
Just curious, should this fix be showing up in 2.0.5 if using WPackagist? I'm including it via Composer in a Bedrock/Sage project, and the `multiple` code in question is commented out. I also tried using `"wpackagist-plugin/cmb2": "dev-trunk"` and same thing, the code is commented out.

I can move to including it locally in the theme to get this fix, but I liked having CMB2 easily updatable as a plugin controlled via Composer.

jtsternberg:
@natebeaty no, we haven't merged to master yet. I expect to soon.

natebeaty:
@jtsternberg ok, thanks for the quick reply. I'll just move CMB2 into my theme for now.

pieroit:
Hi guys,
I need this feature and was happy to see it was already addressed. My problem is that I can't get  the option `multiple` to work with the option `repeatable`.

The meta has two values, sotred into the DB as two different records.
image

If I set `multiple => true` and `repeatable => true`, only the first value is shown in an <input> box and below there is the "Add Row" button to insert more. The second one disappear (despite being present in the DB).
image

If I set `multiple => false` and `repeatable => true`, same as above.

If I set `multiple => true` and `repeatable => false`, both values get passed to the field but they are printed together into one <input>, in which I read `Array`. (see Notice)
image

What do you think? I asked for help before diving into the code because I already have in place a functioning metabox system, and the switch may require too much time. Hope to fix this soon to join the community :)

Thanks,
Piero

pieroit:
bimp :)

pieroit:
Found a way to send the meta values correctly to the repeated field.
This line
https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_Field.php#L232
in the following block:

```
		// If no override, get value normally
		if ( 'cmb2_field_no_override_val' === $data ) {
			$data = 'options-page' === $a['type']
				? cmb2_options( $a['id'] )->get( $a['field_id'] )
				: get_metadata( $a['type'], $a['id'], $a['field_id'], ( $a['single'] || $a['repeat'] ) );
		}
```
calls `get_metadata` with final parameter set to `$a['single'] || $a['repeat']`.
This causes the metabox to load only one value even when you explicitly asked for `repeated` in combination with `multiple`. The problem is fixed by changing to this:
`: get_metadata( $a['type'], $a['id'], $a['field_id'], ( $a['single'] || !$a['repeat'] ) );`

But I don't know if I broke something in this way. Also, the values are still serialized when inserted into the DB. Can anybody help?

Thank you,
Piero

Issue #187, opened 2015-02-13
Repeatable Group: Image gallery empties on changing order


Gaffen:
CMB 2.0.1, Wordpress 4.1

If a repeatable group has an image gallery in it, changing the order removes the gallery from the one that end up on top, replacing it from the gallery below.

Test case:

1. Create two repeatable field items, assign them names and image galleries:
![screen shot 2015-02-13 at 14 42 34](https://cloud.githubusercontent.com/assets/718125/6189126/10479962-b390-11e4-9068-4215b61ed855.png)

2. Update the post

3. Change the order of the items.

4. Update the post

5. One of the galleries disappears and the other is replaced 
![screen shot 2015-02-13 at 14 42 46](https://cloud.githubusercontent.com/assets/718125/6189140/2ce43fee-b390-11e4-8a79-d1a6746778d4.png)

This is inconvenient as it makes it difficult to hand projects over to clients knowing they may accidentally start undoing image galleries.

bonny:
This happens for us too. Re-order = images get lost.

bonny:
Ok, I just made a fix that you can see above. Was going to create a pull request when I saw that this probably was fixed here already (but not merged yet - let's hope it will be soon because we can't release our client website until this is fixed): https://github.com/WebDevStudios/CMB2/pull/153

jtsternberg:
This is fixed in trunk and will be in the next release, thanks to @jamiechong!

Issue #152, opened 2015-01-19
Bug: Sorting Groups with Radio Buttons (With Solution)


jamiechong:
There is a bug when sorting Groups that have radio button input fields. 
They way you sort is by swapping the values of the input fields between the two groups. You're attempting to treat `radio` buttons the same as `checkbox` by doing this in your `cmb.shiftRows` function:

```
			else if ( 'checkbox' === $element.attr('type') || 'radio' === $element.attr('type') ) {
				inputVals[ index ]['$'].prop( 'checked', $element.is(':checked') );
				$element.prop( 'checked', inputVals[ index ]['val'] );
			}
```

Because you're iterating over each input element, you can't swap the `checked` property one-for-one with radio buttons, because when you check a radio button from a group, the dom/browser unchecks the others within the radio button group. Example:

```
Group 1    [ ] A  [X] B   [ ] C
Group 2    [X] A  [ ] B   [ ] C
```

On the first iteration, the `checked` property for A is set in the first group but that unchecks B:
```
Group 1    [X] A   [ ] B   [ ] C
Group 2    [ ] A   [ ] B   [ ] C
```

Now on the second iteration, B is unchecked in Group 1, so B will never get checked for Group 2. 

To solve this, I saved the checked states in a temporary attribute and then set the checked property at the end:

```
		// And swap them all
		$goto.find( cmb.repeatEls ).each( function( index ) {

			...

			// handle checkbox swapping
			else if ( 'checkbox' === $element.attr('type') ) {
				inputVals[ index ]['$'].prop( 'checked', $element.is(':checked') );
				$element.prop( 'checked', inputVals[ index ]['val'] );
			}
			// handle radio swapping
			else if ( 'radio' === $element.attr( 'type' )  ) {
				if ($element.is(':checked')) {
					inputVals[index]['$'].attr('data-checked', 'true');
				}
				if (inputVals[index]['$'].is(':checked')) {
					$element.attr('data-checked', 'true');	
				}
			}

			...

		});

		$parent.find("input[data-checked=true]").prop('checked', true).removeAttr('data-checked');
		$goto.find("input[data-checked=true]").prop('checked', true).removeAttr('data-checked');
```

kynetiv:
+1 this fixes the radio swapping issue for me. Hopeful this can get merged in.

jtsternberg:
related #341 #246

jamiechong:
@jtsternberg  maybe you'll now merge PR #265 ?

Issue #150, opened 2015-01-19
Radio button input isn't saved when adding page to repeatable group


jamiechong:
With the metabox below I have a problem saving the radio button. To repeat the issue:

0. Create a new post
1. Choose "template 1" 
2. Add Page (add Row)
3. Choose "template 2"
4. Publish
5. The 2nd page doesn't even save. It's missing. 
6. Add Page (add Row)
7. Choose "template 2"
8. Update
9. The 2nd page is still missing. 

Note that if I have another field in there, say a text input, the 2nd page will save, but the radio button value doesn't get set. 

Any thoughts on how I can fix this problem? I think it's the root to some other issues I'm having with adding rows/pages and then sorting...

```php
	$meta_boxes[] = array(
		'id'         => CMB_PREFIX.'lb_settings',
		'title'      => 'Settings',
		CMB_POST_TYPE_KEY => array('lookbook'),
		'context'    => 'normal',
		'priority'   => 'high',
		'show_names' => true, 
		'fields'     => array(
			array(
				'id' => CMB_PREFIX.'lb_pages',
				'type' => 'group',
				//'description' => '',
				'options' => array(
					'group_title' => 'Page {#}',
					'add_button' => 'Add Page',
					'remove_button' => 'Remove Page',
					'sortable' => true,
				),
				'fields' => array(
					array(
						'name' => 'Template',
						'id' => CMB_PREFIX.'lb_page_template',
						'type' => 'radio_inline',
						'options' => array(
							'template1' => 'Template 1',
							'template2' => 'Template 2',
							'template3' => 'Template 3',
						)
					),
				)
			),
		)
	);
```

jamiechong:
One thing I just discovered is that when you Add another row/group, the radio button `value` isn't set. 

This is a radio button from a group that saved properly:

    <input type="radio" value="template2" id="_mm_lb_pages_0__mm_lb_page_template2" name="_mm_lb_pages[0][_mm_lb_page_template]" class="cmb2-option">

This is a radio button from the new group:

    <input type="radio" value="" id="_mm_lb_pages_1__mm_lb_page_template2" name="_mm_lb_pages[1][_mm_lb_page_template]" class="cmb2-option" data-iterator="1">


The differences is that the newly added fields don't have a `value` attribute set but they include a `data-iterator="1"` attribute (probably unrelated?).

jamiechong:
[...]

PR #153, opened 2015-01-20
Fixes related to repeatable/sortable groups of fields

jamiechong:
I had some issues with a repeatable and sortable group that contained a radio_inline field and a file_list field. 

The radio button issues were documented in #150  and #152. 

There were a few bugs when trying to sort a group that had a file_list. 

1. There was a javascript error in the loop of `$goto.find( cmb.repeatEls ).each( function( index )` if the number of images in the two file_lists being swapped differed. This was because the `index` id was based off the number of `input` elements, which are hidden for file_lists. I fixed this on line 26 `repeatEls          : 'input:not([type="button"],[id^=filelist]),select,textarea,.cmb2-media-status',`
2. The attribute names of the file_list hidden input fields that set the image id didn't correspond with the correct row number. This is fixed in line `629 to 645`

PR #265, opened 2015-03-26
Fix sorting of Groups with file lists, radio buttons in Groups

jamiechong:
I had some issues with a repeatable and sortable group that contained a radio_inline field and a file_list field. 

The radio button issues were documented in #150  and #152. 

There were a few bugs when trying to sort a group that had a file_list. #187 

1. There was a javascript error in the loop of `$goto.find( cmb.repeatEls ).each( function( index )` if the number of images in the two file_lists being swapped differed. This was because the `index` id was based off the number of `input` elements, which are hidden for file_lists. I fixed this on line 26 `repeatEls          : 'input:not([type="button"],[id^=filelist]),select,textarea,.cmb2-media-status',`
2. The attribute names of the file_list hidden input fields that set the image id didn't correspond with the correct row number. This is fixed in line `629 to 645`


jamiechong:
I'm not sure what the "Failed" check is, but this pull request is legit. It's the same PR as #153 but you wanted me to apply changes to `trunk`. Hopefully you can get this into a near-future release! At least look at the changes - maybe you want to implement the fixes in another way. Thanks!

jamiechong:
trashing this PR.

PR #276, opened 2015-04-08
Update VERSION constant

pmgarman:
Version 2.0.5 is 2.0.5 right? Since the class is *_205_trunk, should the VERSION constant be '2.0.5-trunk' also? or should this actually be 2.0.6-trunk since 2.0.5 is already tagged?


jtsternberg:
hmm it's actually wrong on master as well. Your patch is right though. Thank you.

PR #303, opened 2015-04-28
Fix issue with on_front or show_on_cb

sprightly:
If have one form in backend, and with help of on_front or show_on_cb, display it partially at frontend, the fields\meta that not displayed, and not exist in submission, are removed during submit process. For group field exist [proper check](https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2.php#L481), but usual fields not have some, that is fixed by PR.


jtsternberg:
Only checking for `null` will definitely break things. It would be much better to check `$this->args( 'on_front' )`.

Issue #257, opened 2015-03-18
Adding a group row duplicates the previous row's content


austinjreilly:
It's entirely possible that I'm doing something that isn't smart, so if that's the case please let me know. In any case, here's what I've got. The relevant portions of my code are below.

I am using CMB2 to create an internal "Notes" field that will add the current user id as the note author, along with the content of the note. Then on refresh, the note author will display above the note (via the `get_note_author` callback). This works with the code I have below, except that now when I click the "Add a note" button, the new note contains the old note's content and author ID.

It looks to my untrained eye like something that I'm doing is preventing the `newRowHousekeeping` JavaScript function from firing, but I'm definitely not sure.

Anyone have any ideas?

```php
//Create notes metabox
add_filter('cmb2_init', array($this, 'notes_metabox'));

//Renders note author id
add_filter( 'cmb2_render_note_author', array($this,'cmb2_render_callback_for_note_author'), 10, 1 );

//Sanitizes note author id field
add_filter( 'cmb2_sanitize_note_author', array($this,'cmb2_sanitize_note_author_callback'),10, 1);
```

```php
public function notes_metabox() {

    $prefix = '_journal_notes_';

    $journal_notes_metabox = new_cmb2_box(array(
        'id' => $prefix . 'metabox',
        'title' => 'Notes',
        'object_types' => array('journal'),
        'context' => 'side',
        'priority' => 'default',
        'show_names' => false,
        'sortable' => false,
        //'closed' => true,
        'fields' => array(
            array(
                'id' => $prefix . 'repeat_group',
                'type' => 'group',
                'options' => array(
                                'group_title' => 'Note #{#}',
                                'add_button' => 'Add a note',
                                'remove_button' => 'Remove this note',
                             ),
                'fields' => array(
                                array(
                                    'name' => 'Note',
                                    'desc' => '',
                                    'id' => 'content',
                                    'before_row' => array($this,'get_note_author'),
                                    'type' => 'textarea_small',
                                    'default' => false
                                ),
                                array(
                                    'name' => 'Note Author',
                                    'desc' => '',
                                    'id' => 'author',
                                    'type' => 'note_author',
                                    'default' => ''
                                ),
                            ),
            ),
        ),
    ));
}
```

```php
//Adds hidden note author field
public function cmb2_render_callback_for_note_author($field_type_object){
    echo $field_type_object->input( array( 'class' => 'cmb2_note_author', 'type' => 'hidden' ) );
}

[...]

Issue #263, opened 2015-03-24
Group Radio Input Values


f-o-n:
Hi,
something about groups again: 
I have the issue that the newly dynamically generated group rows don't have any values assigned to the radio inputs.
The values only appear if I save my post. I use the radio input to show/hide other fields and consistent radio input values would simplify this task greatly.

Awesome plugin btw. Much obliged.

jtsternberg:
Do you have a screenshot and/or code? Help us help you. :)

f-o-n:
Yes sure, here are some snippets that I tested in a fresh WordPress install.

I have a Post Type to create Slideshows. I created a checkbox so the user can easily switch between Images and Videos so he doesn't has both options. The code works if I press save once after creating the repeatable groups and the input gets an value assigned. 


Before saving the post:
![presave](https://cloud.githubusercontent.com/assets/11637643/6997260/f91c20fc-dbb5-11e4-8167-c3419cb846c8.png)

After saving the post:
![postsave](https://cloud.githubusercontent.com/assets/11637643/6997261/05fe402a-dbb6-11e4-9da2-014ee0765d0c.png)


Here is my MetaBox code:

```php
function my_gallery_group_metabox() {
    $prefix = '_my_gallery_';

    $cmb_group = new_cmb2_box( array(
        'id'           => $prefix . 'group_metabox',
        'title'        => __( 'Gallery Slides', 'my-gallery' ),
        'object_types' => array('gallery'),
    ) );

    $group_field_id = $cmb_group->add_field( array(
        'id'          => $prefix . 'group',
        'type'        => 'group',
        'options'     => array(
            'group_title'   => __( 'Slide {#}', 'my-gallery' ),
            'add_button'    => __( 'Add Slide', 'my-gallery' ),
            'remove_button' => __( 'Remove Slide', 'my-gallery' ),
            'sortable'      => true,
        ),
    ) );

    $cmb_group->add_group_field( $group_field_id, array(
        'name' => __( 'Slide Title', 'my-gallery' ),
        'id'   => 'title',
        'type' => 'text_medium',
    ));
    
    $cmb_group->add_group_field( $group_field_id, array(
        'name'    => __( 'Slide Type', 'my-gallery' ),
        'id'      => 'type',
        'type'    => 'radio_inline',
        'options' => array(
            'image' => __( 'Image', 'my-gallery' ),
            'video'   => __( 'Video', 'my-gallery' ),
        ),
[...]

Issue #246, opened 2015-03-11
Sortable group not saving radio options initially


joshuadavidnelson:
I've created a sortable group with a couple of fields, one of which is a radio option that defaults to one of three choices. It appears to work just fine in populating new rows to the group, and saving values from the other fields, but on the post-new.php save it only stores the first group's radio button value and the others revert to the default. It works fine on the the post.php screen's save, so it just needs to be created and then re-saved with new values.

jtsternberg:
related #341

jtsternberg:
Should be good in the trunk branch. please test.

joshuadavidnelson:
This looks to be working now, thanks!

PR #304, opened 2015-04-28
Fix issues related to group repeatable subsequent rows when there is a checkbox or radio input.

kynetiv:
If there is a checkbox or radio input in the Group, the original value is needed or saving the post will not save the checkbox or radio value (subsequent post saves work because inputs will then have the required value set).
This patch conditionally adds values to radios and inputs on new repeatable rows and sets the checked attribute only on the default option (radios).
Fixes issue #257 (this will reset text inputs as well), and #246 & #263(restores radio and checkbox values on new row saves)


kynetiv:
Well, need to figure out how these tests are working so it can pass, but also found that there is a bug to fix when there are multiple radio options on the second subsequent new row. Will update this in a bit.

jtsternberg:
Should be good in the trunk branch. please test.

tw2113:
Closing as it is said to be covered already in trunk, and probably release versions now.

Issue #333, opened 2015-05-27
Uncaught TypeError: $id(...).hide(...).farbtastic is not a function


jedrzejchalubek:
Getting error form js file: `Uncaught TypeError: $id(...).hide(...).farbtastic is not a function` and of course all js based repeatable UI elements dont work.

Wordpress: 4.2.2
CMB: dev-trunk

jtsternberg:
The only reason that would happen is if this condition: `(typeof jQuery.wp === 'object' && typeof jQuery.wp.wpColorPicker === 'function')` is false. This means the colorpicker is not enqueued or some other script is causing that script to not work. There is nothing in CMB2 that would cause this to occur.

jedrzejchalubek:
But on version CMB2 2.0.6 this error not appears.

jtsternberg:
If you're replacing default WordPress scripts, there's not much I can do to help.

michaeljpowell:
I get this same error when I'm on a custom options page (specifically with the file upload field). When it's within a page it appears to be working fine.

I am using the WP plugin version.

EvanHerman:
I too was getting the same error using the version up here on GitHub and the version found in the WordPress plugin repo. Whenever I included a color picker field I was getting the same error as OP.

This was on a fresh TwentyFifteen install with no other plugins installed.

Adding the following, just below colorpicker field definition, did the trick for me:
```php
wp_enqueue_style( 'wp-color-picker' );
wp_enqueue_script( 'wp-color-picker' );
```

It seems to only be the color picker field that's causing this issue.


Evan

jtsternberg:
Thanks for reporting. The issue should be addressed in https://github.com/WebDevStudios/CMB2/commit/244d39fc3c27cfc195e495ecb331fac73e51399a

PR #336, opened 2015-05-31
Enqueue color picker script

EvanHerman:
The color picker script seem to be missing. When using a color picker field type a javascript error is thrown, and causing a few styles and other JavaScript functionality to not load/function properly.

I've tested this on a fresh install of TwnetyFifteen with no other plugins installed/active.

I tested with both the WordPress.org repo version and the version up here on Github. After enqueueing the color picker scripts manually, everything worked fine.

<strong>Edit:</strong>
After doing a little more work, I've displayed the user profile fields on the front end of a site using:
```php
cmb2_metabox_form( '_user_profile_edit', get_current_user_id() );
```

It looks like the file for cmb2-front.min.css didn't make it into the release. I was getting a 404. After re-compiling and including, things seem ok.

Related: Issue #333


jtsternberg:
The dependency issue is addressed in https://github.com/WebDevStudios/CMB2/commit/244d39fc3c27cfc195e495ecb331fac73e51399a

EvanHerman:
Great, thanks for addressing the issue.

Evan

Issue #357, opened 2015-06-20
File_List field type does not have expected multi-select behaviour


SteveHoneyNZ:
Essentially the 'isList' Boolean check [here](https://github.com/WebDevStudios/CMB2/blob/master/js/cmb2.js#L192) results in the wp.media frame created always being in a 'single select' state, whether called by a `file_list` or `file` field type.

Re-writing as ```multiple: isList ? 'add' : false ``` results in the `file_list` media frame allowing multi-select as I would expect.

I get somewhat confused looking at the WordPress core functions around this, in [media-views.js](https://github.com/WordPress/WordPress/blob/master/wp-includes/js/media-views.js) there are various references to the multiple option being both Boolean and supporting 'add' or 'toggle' string values. Bit beyond me TBH, a little more background on this [SE question](http://stackoverflow.com/questions/17473885/wordpress-3-5-media-uploader-multiple-file-select).

The `upload_file` language string should also be updated to reflect being used on both single and multiple uploads, currently `Use this file` for both, simplest would be `Insert File(s)`, rather than a whole new language string being added and checked for on the `button:` option. "Insert" also follows  the common syntax of WordPress rather than "Use" or "Add".

PR #358, opened 2015-06-20
Fix for Issue #357

SteveHoneyNZ:
Changes as outlined in issue #357


jtsternberg:
Thanks! I opted to include a separate string for multiple, and am sticking with the 'use' language as it is technically more accurate since you are not inserting the files to the post-content.

PR #360, opened 2015-06-22
less fuzzy search for image attachment ID

### LINKED ISSUE #137: image-Field saves right URL, but wrong ID
A image-Field inside a group-Field:

```php
array(
	'id'          => $prefix . 'referenz',
	'type'        => 'group',	
	'options'     => array(
		'add_button'    => __( 'Add Image', 'huku' ),
		'remove_button' => __( 'Remove Image', 'huku' ),
		'sortable'      => true,
	),	
	'fields'      => array(						
		array(
			'name' => 'Image',
			'id'   => 'image',
			'type' => 'file',
		),					
	),
),
```
This works in most cases, but not in this case:
(all following URLs are shortened)

I add 'uploads/2015/01/spitz.png' and it shows me the right Preview-Image.
When i save the post, it shows me the right URL ('uploads/2015/01/spitz.png'),
but the wrong Preview-Image -> 'uploads/2014/12/logo_spitz.png'.

The Database-Records holds also the right URL and the wrong ID (the ID of 'uploads/2014/12/logo_spitz.png' ).
I suppose it has something todo with the similar filename, both ends with spitz.png.

I give you more Details if needed.

cbergen:
Fixes #137 and also the case where an image url changes due to editing in the media library.

### DIFF
--- a/includes/CMB2_Utils.php
+++ b/includes/CMB2_Utils.php
@@ -28,17 +28,20 @@ class CMB2_Utils {
 	public function image_id_from_url( $img_url ) {
 		global $wpdb;
 
-		$img_url = esc_url_raw( $img_url );
-		// Get just the file name
-		if ( false !== strpos( $img_url, '/' ) ) {
-			$explode = explode( '/', $img_url );
-			$img_url = end( $explode );
+		$new_path = esc_url_raw($img_url);
+		
+		// Strip the baseurl from the full image url 
+		$uploads = wp_upload_dir();
+		if ( 0 === strpos( $new_path, $uploads['baseurl'] ) ) 
+		{
+			$new_path = str_replace( $uploads['baseurl'], '', $new_path );
+			$new_path = ltrim( $new_path, '/' );
 		}
+		
+		// Get the attachment ID from the postmeta table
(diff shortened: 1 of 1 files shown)


tw2113:
@cbergen I realize it's been a year, but by chance can you recall an example usecase for this and what it helps take care of?

cbergen:
It's been awhile. And looking at this again I don't think my PR is the best solution.

The use case is explained in issue #137. The image_id_from_url() function uses a LIKE select query to find the attachment ID. This is error-prone as it is common for there to be images with similar names.  

My change uses the postmeta key: _wp_attached_file. This is the same database location [get_attached_file()](https://developer.wordpress.org/reference/functions/get_attached_file/) uses.

But my solution assumes that the url value in _wp_attached_file is always relative and judging from the implementation of get_attached_file() it's not.

jtsternberg:
Closing, but feel free to open another if you encounter a better solution. Also, please open any new PRs against the trunk branch (as per [contribution guidelines](https://github.com/WebDevStudios/CMB2/blob/master/CONTRIBUTING.md#how-to-contribute-to-cmb2)).

PR #361, opened 2015-06-22
Fix 'select' options defined as arrays.

tommusrhodus:
If a select option was defined in the old CMB style of:

    'options' => array(
       array('name' => 'Normal Width, Normal Height (285x190px)', 'value' => 'normal'),
    )

The result was broken. The above adds in back compatibility for this formatting style (old CMB).


jtsternberg:
We actually removed that compatibility layer a couple versions ago as CMB2 is not really meant to be compatible w/ the original, and it requires more upkeep.

PR #367, opened 2015-06-28
Fix output the multicheck_inline field type

ediamin:
The function multicheck_inline in CMB2_Types.php must return the html element. Otherwise _render function doesn't echo anything.


jtsternberg:
Good catch, and thank you!

PR #371, opened 2015-07-01
Add quotes to prevent globbing for test set up bin script

bradp:
also replace need for wgetting external resource


bradp:
closed for #372

PR #380, opened 2015-07-08
Fix German translation of time format string (and a typo).

superfreund:


jtsternberg:
Hello @superfreund, thank you for your contribution. We're handling all translation through http://wp-translations.org/cmb2, can you submit the update there?

PR #381, opened 2015-07-08
Optimize images

bradp:


bradp:
Ran everything through imageoptim / testing travis notifications

bradp:
Not worth it - we lose more space in git history than the actual image sizes

PR #382, opened 2015-07-09
Define $formfield when mediaHandler fires instead of closure

augustuswm:
cmb.mediaHandlers.* are redefined on the first click of each “file” or “file_list” on the page. Each time .list and .single are redefined the local variables for the current “file” and “file_list” are closed over. This breaks the handlers for any previously clicked “files” and “file_lists”. Variables have been moved into the handlers themselves rather than rewriting the handlers.

It might be more elegant to define new handlers for each "file" or "file_list" that still use the closure over $formfield in particular as opposed to the approach here.

In reference to issues: https://github.com/WebDevStudios/CMB2/issues/250, https://github.com/WebDevStudios/CMB2/issues/296


augustuswm:
Following up to see if there is anything I can do to assist in getting either this fix added or implementing a fix for the issues in another way.

jtsternberg:
@augustuswm sounds like you have a good handle on a possible solution. Would you be willing to submit a PR?

jtsternberg:
Wow, that's embarrassing.. just realized this IS a PR. ha ha.. I think I have it sorted though.

jtsternberg:
These commits should take care of the issue: https://github.com/WebDevStudios/CMB2/compare/6db15bdd147d2c041bf5ffcb5333e2702d1f7456...7ca260651873478f124675b18f85b2fc2d2d45b9

PR #387, opened 2015-07-15
Fix delete row from repeatable group. Fixes #281

### LINKED ISSUE #281: Can't delete row from repeatable group
To delete a row from a repeatable group, I find I first have to move it to the bottom of all rows.

Or more accurately, I delete the row (by either the X or "remove" button) and update the post and it's still there. By then moving this "phantom" row down so it's the last entry and updating again, it disappears on update (without a further delete command).

### LINKED ISSUE #357: File_List field type does not have expected multi-select behaviour

yuks:
Fixes #281

### DIFF
--- a/js/cmb2.js
+++ b/js/cmb2.js
@@ -467,19 +467,17 @@ window.CMB2 = (function(window, document, $, undefined){
 		var $this = $(this);
 		var name  = $this.attr( 'name' ); // get current name
 
-		// No name? bail
-		if ( typeof name === 'undefined' ) {
-			return false;
+		// If name is defined
+		if ( typeof name !== 'undefined' ) {
+			var prevNum = parseInt( $this.parents( '.cmb-repeatable-grouping' ).data( 'iterator' ) );
+			var newNum  = prevNum - 1; // Subtract 1 to get new iterator number
+	
+			// Update field name attributes so data is not orphaned when a row is removed and post is saved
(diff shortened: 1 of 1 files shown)


jtsternberg:
Thank you!

jtsternberg:
Oops, I just realized I merged your PR to the master branch which is our release branch. In the future, can you please submit PRs to the trunk branch?

yuks:
Sorry. Next time, will submit to the trunk branch.

PR #389, opened 2015-07-15
Returning String type

vajrasar:
returning `string` type `$string` when condition enters `else` during the creation of valid timestamp to save in the database.


jtsternberg:
Thank you!

PR #401, opened 2015-07-22
load_theme_textdomain expects an absolute path

### LINKED ISSUE #400: Windows IIS Fatal Error on Go Daddy
Using WordPress on GoDaddy's IIS and installing CMB2 results in the following error:

`PHP Warning:  is_readable(): open_basedir restriction in effect. File(/languages/en_US.mo) is not within the allowed path(s): (G:/PleskVhosts//give.advancedwp.org\;C:\Windows\Temp\) in G:\PleskVhosts\give.advancedwp.org\httpdocs\wordpress\wp-includes\l10n.php on line 476`

Which basically means that the server configuration has a setting for permitted file paths, and that this plugin does not match the required pattern... hence the error. 

I've narrowed it down to the `l10ni18n` method: https://github.com/WebDevStudios/CMB2/blob/master/init.php#L157

I'm looking into a fix right now that'll make IIS happy and preserve the method's functionality.

DevinWalker:
Fixes #400

### DIFF
--- a/init.php
+++ b/init.php
@@ -160,7 +160,7 @@ if ( ! class_exists( 'CMB2_Bootstrap_209_Trunk', false ) ) {
 				$loaded = load_muplugin_textdomain( 'cmb2', '/languages/' );
 			}
 			if ( ! $loaded ) {
-				$loaded = load_theme_textdomain( 'cmb2', '/languages/' );
+				$loaded = load_theme_textdomain( 'cmb2', get_template_directory() . '/languages/' );
 			}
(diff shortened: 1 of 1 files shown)


jtsternberg:
This should probably use `get_stylesheet_directory` to account for child-theme usage.

DevinWalker:
Good point...

DevinWalker:
Good to go now

jtsternberg:
Thank you!

PR #430, opened 2015-08-13
Add filter for non-repeatable fields

JPry:
This also streamlines how the field type is checked. Using `isset()` is more efficient than `in_array()`.

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -519,18 +519,33 @@ class CMB2_Field {
 	 */
 	public function repeatable_exception( $type ) {
 		// These types cannot be escaped
-		return in_array( $type, array(
-			'file', // Use file_list
-			'radio',
-			'title',
+		$internal_fields = array(
+			// Use file_list instead
+			'file'                => 1,
+			'radio'               => 1,
+			'title'               => 1,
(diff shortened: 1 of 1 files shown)


jtsternberg:
I'm not sure I understand the value of reversing this to use array keys?

JPry:
@jtsternberg 

> I'm not sure I understand the value of reversing this to use array keys?

Here's a [StackOverflow](http://stackoverflow.com/questions/13483219/what-is-faster-in-array-or-isset?answertab=votes#tab-top) answer that should explain it.

jtsternberg:
I was familiar w/ the fact that isset is faster than in_array, but having a hard time believing the value gained from using array keys (for the number of items that would live in this array), is worth developer confusion w/ using this filter with array keys. 
Convince me. ;)

JPry:
@jtsternberg It's true that any single instance of this method being called won't produce any significant performance increase, and it may add some confusion. However, this method is called every single time an instance of `CMB2_Field` is created. There are 5 calls currently in `CMB2`, and a single call each in `CMB2_Sanitize` and `CMB2_Types`.

Taking into account multiple fields and multiple metaboxes, this very quickly increases by a significant factor. Thus, the performance improvement can add up and be worth the slight confusion developers may have.

jtsternberg:
Color me convinced. Thank you for thinking this through.

Issue #348, opened 2015-06-05
Repeating Groups with Repeating Text Fields


bearded-avenger:
We lean on CMB2 to create quizzes. These quizzes have questions (repeatable groups). Within each question, the admin can create answers (repeatable text field). Here's an example of one of the repeatable groups structure.

![image 2015-06-05 at 7 58 48 am](https://cloud.githubusercontent.com/assets/1317303/8007129/7499ca4e-0b5f-11e5-9123-7bc13239d49a.png)

When the data is saved, it removes all but one of the questions. To replicate:
- First group field - add 3 repeating text fields 
- click "add group" - add 3 repeating text fields
- do this two more times to make a total of 5 repeating field groups, with each group having 3 repeating text fields
- publish the post

When you refresh the page:
- repeating group fields 2,3,4 now only have 2 repeating text fields (it removed the last)
- it added a 6th group with two random answers

Here's the sample code:
```php
		// instantiate metabox
	    $cmb = new_cmb2_box( array(
	        'id'            => '_quiz_metabox',
	        'title'         => 'Quiz Questions',
	        'object_types'  => array( 'cgc_quiz' )
	    ) );

	    $cmb->add_field( array(
	        'name'       => 'Public Quiz',
	        'id'         => '_is_public',
	        'type'       => 'checkbox'
	    ) );

	    $cmb->add_field( array(
	        'name'       => 'XP Worth',
	        'id'         => '_question_xp',
	        'type'       => 'text_small'
	    ) );

	    $cmb->add_field( array(
	        'name'       => 'Questions to Display',
	        'desc'       => 'Image URL',
	        'id'         => '_total_questions',
	        'type'       => 'select',
	        'default'		=> '4',
			'options' 		=> array(
		        '4' 	=> '4',
		        '6' 	=> '6',
		        '8' 	=> '8'
			),
	    ) );

		// $group_field_id is the field id string, so in this case: $prefix . 'demo'
		$group_id = $cmb->add_field( array(
			'id'          		=> '_quiz_questions',
			'type'        		=> 'group',
			'description' 		=> 'A Question with Answers',
			'options'     		=> array(
				'group_title'   => 'Question {#}',
				'add_button'    => 'Add Another Question',
				'remove_button' => 'Remove Question',
				'sortable'      => true
			)
		) );

		$cmb->add_group_field( $group_id, array(
			'id'			=> 'question',
			'name' 			=> __('Question', 'cgc-quiz'),
[...]

PR #433, opened 2015-08-20
fixes #187 where sorting filelists don't save in repeatable groups

### LINKED ISSUE #187: Repeatable Group: Image gallery empties on changing order

jamiechong:
Can't wait for these changes in the Wordpress.org version!

### DIFF
--- a/js/cmb2.js
+++ b/js/cmb2.js
@@ -18,7 +18,7 @@ window.CMB2 = (function(window, document, $, undefined){
 	// CMB2 functionality object
 	var cmb = {
 		idNumber        : false,
-		repeatEls       : 'input:not([type="button"]),select,textarea,.cmb2-media-status',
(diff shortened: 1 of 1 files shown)


jamiechong:
I have a feeling this fix could probably be refactored a bit to also solve #348.

jamiechong:
@jtsternberg  what do you think? Good enough to merge into trunk? Or should I investigate 348 and create a PR for both?

Issue #330, opened 2015-05-23
text_datetime_timestamp as a repeater field not saving proper timestamp


tw2113:
It's saving the current time instead of the set time.

https://wordpress.org/support/topic/date-will-not-save-saves-today-instead?replies=4#post-6984530

tw2113:
@jtsternberg have you looked into this one at all? Original reporter is still having issues and wondering if there's possible chance for some sort of fix for it.

turegjorup:
I can confirm I'm seeing the same issue. WP 4.4 beta4, CMB2 2.1.2 on both Firefox and Chrome. I'm also only seeing it with a repeatable datetime field. 

Looking through #446 there are some changes to the code taht fails, but I still don't think they properly account for repeating datetime fields.

jtsternberg:
This should be good in [trunk](https://github.com/WebDevStudios/CMB2/tree/trunk). Please test.

turegjorup:
Just tested with WP 4.4-beta4-35734, repeatable datetime fields now works as expected. Thank you for fixing it.

PR #446, opened 2015-08-29
fix so datetime combo field respects date-format and time format a set up by the user

yivi:
I still have to review the datetime-timezone field.

A couple things:
Since in CMB_Field.php:630 you have 
`return date( stripslashes( $this->args( $format ) ), $meta_value );`
users wont be able to use escape characters in their format strings. (E.g.: 'd \d\e m \d\e Y', or any other combinations).

Either that should be reflected in the docs, or that particular stripslashes should go... :P

I edited the wiki to reflect that there are no longer separate "text_date" and "text_date_timestamp". But I guess we should also put there the acceptable characters for the format string (d, j, z, m, n, y, and Y, instead of the whole date() range).

I'll try to finish up the timezone combo later today.

Regards!


yivi:
Timezone combo done as well.
As an aside... for date/time combo (no timezone) we are saving an array with 'date' and 'time' as keys, while just saving a timestamp should be sufficient (and maybe cleaner?)

Regards, hope this helps.

yivi:
@jtsternberg, this PR has current trunk merged in, and I have tested it and have it working for all date/time types I'm aware of.

includes/CMB2_Sanitize.php
```diff
@@ -212,10 +225,12 @@ public function text_datetime_timestamp( $repeat = false ) {
 			return $repeat_value;
 		}
 
-		$this->value = strtotime( $this->value['date'] . ' ' . $this->value['time'] );
+		if ( is_array( $this->value ) && isset( $this->value['date'] ) ) {
```
turegjorup: Does this account for repeating datetime fields? 

I'm seeing this issue mentioned in #330 as well. I was looking through the code and believe I have narrowed it down to the deleted line 215 failing to account for repeatable fields. However I havnt had the time to try to fix it.

jtsternberg:
Thanks for all your hard work here. I've merged it in and have made some changes to reign the scope in a bit. It seems to work well in my testing, but please test on your end to verify.

> users wont be able to use escape characters in their format strings. (E.g.: 'd \d\e m \d\e Y', or any other combinations).

> Either that should be reflected in the docs, or that particular stripslashes should go... :P

What is the use-case where that is necessary?

> But I guess we should also put there the acceptable characters for the format string (d, j, z, m, n, y, and Y, instead of the whole date() range).

Good idea, I updated the trunk branch of the wiki and will merge it to master when the next version of CMB2 is released with this update/change.
`
> As an aside... for date/time combo (no timezone) we are saving an array with 'date' and 'time' as keys, while just saving a timestamp should be sufficient (and maybe cleaner?)

I saw that in your branch, but that is not part of how CMB2 currently works. I've removed it and it's using the timestamp as it should.

includes/CMB2_Sanitize.php
jtsternberg: I'll look into the repeatable field issues (hopefully tomorrow).

PR #453, opened 2015-09-09
Update CMB2_Utils.php

wpscholar:
After having done a decent amount of research and testing on properly fetching an image ID from a URL, I've found this solution to be more robust.  Ultimately, using the guid just isn't reliable enough: http://wpscholar.com/blog/get-attachment-id-from-wp-image-url/

### DIFF
--- a/includes/CMB2_Utils.php
+++ b/includes/CMB2_Utils.php
@@ -23,13 +23,13 @@ class CMB2_Utils {
 	 * Utility method that attempts to get an attachment's ID by it's url
 	 * @since  1.0.0
 	 * @param  string  $img_url Attachment url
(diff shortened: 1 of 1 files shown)


jtsternberg:
This looks good. The image_id_from_url method is actually a backup method in case the image ID didn't get properly saved to the hidden field. Are you seeing it fall back to this method often?

wpscholar:
@jtsternberg I was actually just perusing the code and noticed this.  Knowing that the guid isn't the most reliable way, I figured I would throw in my two cents.

jtsternberg:
Had to fix my broken tests before I could merge your changes in, but got them in. This is merged into trunk: https://github.com/WebDevStudios/CMB2/commits/trunk

PR #454, opened 2015-09-12
Update CMB2_Sanitize.php

superfreund:
Value-changes in 'taxonomy_radio_inline'- and 'taxonomy_multicheck_inline'-fields are not being updated unless sent through default_sanitization-function.


jtsternberg:
Good catch, thank you.

PR #460, opened 2015-09-16
Set terms instead of meta for taxonomy_multicheck_inline

JacobDorman:
missing from case statement, so taxonomy_multicheck_inline saves to `wp_postmeta` instead of `wp_term_relationships`


jtsternberg:
Thank you for your contribution. This has already been addressed in 2.1.1

PR #463, opened 2015-09-20
WP_DEVELOP_DIR != $test_root

quasel:
As far as i know WP_DEVELOP_DIR Stands not for the directory of the Tests but CMB2 sees it that way ;) 
'WP_DEVELOP_DIR' != 'WP_TESTS_DIR'

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -12,7 +12,7 @@
 // 2. Plugin installed inside of WordPress.org developer checkout
 // 3. Tests checked out to /tmp
 if ( false !== getenv( 'WP_DEVELOP_DIR' ) ) {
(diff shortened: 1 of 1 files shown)


quasel:
Did a bit of research seems Jetpack uses WP_DEVELOP_DIR this way - many other seem to use WP_TESTS_DIR without looking at WP_DEVELOP_DIR or like pods just adding . '/tests/phpunit'

Makes the most sense to me at least from reading the Words ;) develop dir ... 

Problem is this might break some testing setups - maybe

quasel:
Another one using pods approach WP-API 
https://github.com/WP-API/WP-API/blob/develop/tests/bootstrap.php#L21

jrfnl:
+1

PR #473, opened 2015-10-01
Add field specific save_field filter

wpsmith:


jtsternberg:
Requested a cleaned up, documented version

PR #474, opened 2015-10-01
Add an action to fire after the field has been saved/updated.

wpsmith:
Prefer an action over filtering $updated to ensure that updated returns properly providing a hook. This will have a minimal impact.


wpsmith:
Moved to #474

PR #475, opened 2015-10-01
Save Field Hook

wpsmith:


jtsternberg:
Made a few small tweaks and added unit tests.

PR #476, opened 2015-10-01
taxonomy_select none option fix

adamcapriola:
I have a taxonomy with numeric terms — 0, 1, 2, 3, etc... — and the `taxonomy_select` `option` output was looking like this:

```html
<option value="" selected="selected">None</option>
<option value="0" selected="selected">0</option>
<option value="1">1</option>
```

Doing a strict comparison seems to fix the issue and only `None` gets selected. I can then select the `0` value and it is saved and selected correctly too.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -682,7 +682,7 @@ class CMB2_Types {
 			$options .= $this->select_option( array(
 				'label'   => $term->name,
 				'value'   => $term->slug,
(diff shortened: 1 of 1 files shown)


jtsternberg:
Please open this pull request against the trunk branch. (and please review the [CONTRIBUTING.md](https://github.com/WebDevStudios/CMB2/blob/master/CONTRIBUTING.md))

adamcapriola:
Oops! My bad.

Issue #158, opened 2015-01-25
Bug: Repeatable oEmbed breaks form


GregLancaster71:
This is a new bug, this latest revision of the plugin really broke my plugin and website good. Repeatable oEmbeds return an array in the input field.  

Screenshot: http://snag.gy/zMXoi.jpg

Error: `Notice: Array to string conversion in /Applications/AMPPS/www/local.dev/wp-content/plugins/simple_review_final/CMB2/includes/CMB2_Types.php on line 187`

It appears to me [for me] that all repeatable text fields are broken.

jtsternberg:
@GregLancaster71 last update should fix that. Please let me know.

fpswebmaster:
I'm experiencing this issue as well (using the newest release 2.02) when upgrading from CMB 1.2.0 to 2.0.2 All Repeater Fields [text] are filled with the term "Array" and the error "Array to String Conversion on...".

This only happens if the field is empty, almost as if the initiation of the meta field on the given post is being stored as an array.

jtsternberg:
CMB2 is not 100% back-compatible with CMB1 (which is why they are different libraries). I can't recall how 1.2 did repeat fields vs how they are now, but it's possible you may need a script to convert the previous values to the new structure.

fpswebmaster:
Understood, after a simple SQL script I was able to clean it out. And to be honest because it was just injecting "array()" into the blank fields, it wasn't effecting the functionality of the website since thats considered blank (for all of my IF (empty($VAR)) statement)

Thank you for the quick reply

PR #487, opened 2015-10-08
Issue #158

whitesunset:
Fix error: Notice: Array to string conversion in /CMB2/includes/CMB2_Types.php on line 187

Bug is still alive on trunk
Steps to reproduce:

1. Create field with type = group
2. Add field with type = text to group. 

For example: http://pastebin.com/bRTAE4pR


whitesunset:
First try here https://github.com/WebDevStudios/CMB2/pull/485

jtsternberg:
I cannot duplicate the issue with your exact code.

tw2113:
Going to close as "worksforme". If you can reproduce with an uptodate version of trunk, we can re-investigate.

Thanks.

PR #491, opened 2015-10-12
Feature/save field filter

wpsmith:
Change filter name & add $new_value to filter


wpsmith:
See #475

jtsternberg:
`field_id` is what is used in nearly every CMB2_Field, the field type can be retrieved from the callback via `$field->type()`, and the (proper and filtered) new value can be be retrieved from the callback via `$field->value` (I just ensured `$field->value` would be up-to-date when these actions are called: https://github.com/WebDevStudios/CMB2/commit/1958844f4b055ad134e6a7dadacc54b3a1b86e55).

PR #318, opened 2015-05-08
datepicker and timepicker fixes

yivi:
Support for `date-format` and `time-format` in text_date and text_time fields. They accept a php formatted string, which is converted to a js formatted string and injected in the data attribute so the pickers can be initialized appropriately.  Fixes #282, fixes #300

PR #498, opened 2015-10-19
Support for custom "date_format" and "time_format" arguments

mweimerskirch:
Support for custom "date_format" and "time_format" arguments for the text_date, text_date_timestamp and text_datetime_timestamp fields.

Original fixes by @yivi, based on https://github.com/WebDevStudios/CMB2/pull/318

I removed all the unrelated changes, and added support for the time picker. To make merging easier, the pull request also doesn't include a minified JS file.

### DIFF
--- a/includes/CMB2_Sanitize.php
+++ b/includes/CMB2_Sanitize.php
@@ -209,7 +209,19 @@ class CMB2_Sanitize {
 	 * @return string       Timestring
 	 */
 	public function text_date_timestamp() {
-		return is_array( $this->value ) ? array_map( 'strtotime', $this->value ) : strtotime( $this->value );
+		if ( is_array( $this->value ) ) {
+			$returnValue = [ ];
+			foreach ( $this->value as $value ) {
+				$date_object   = DateTime::createFromFormat( $this->field->args['date_format'], $value );
+				$returnValue[] = $date_object ? $date_object->setTime( 0, 0, 0 )->getTimeStamp() : '';
+
+			}
+		} else {
+			$date_object = DateTime::createFromFormat( $this->field->args['date_format'], $this->value );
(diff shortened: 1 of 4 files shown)


jtsternberg:
Why do you believe your PR is a better option than @yivi's #318? What do you consider to be "unrelated changes"?

mweimerskirch:
I downloaded the patch file for #318 directly from GitHub because I didn't want to pull in 23 individual commits and there were hundreds of whitespace changes as well as changes in a translation file: https://github.com/WebDevStudios/CMB2/pull/318.patch
But just now after reading your message I noticed that those changes are not visible when viewing the diff online. Seems to be a bug on GitHub, so I take the part about the "unrelated changes" back.

yivi:
yivi here.

I see that in your version you are are back to using DateTime::createFromFormat, whereas we had a workaround for PHP versions < 5.3.

Also, my work continued in #446, on a different branch that @jtsternberg created to carry on this work. I updated it a couple weeks ago, and it does have support for time_picker as well...

I think I messed up with the PRs, by opening a new one instead of continuing work in #318, but I can merge trunk into #446 again to see if everything keeps working as intended.

Regards.

jtsternberg:
@yivi we're so close. Thanks for your efforts. Let me know which (#318 or #446) I should be testing against. For now, i've merged trunk into the WebDevStudios:yivi-date-time-picker-fixes branch so it is up-to-date.

mweimerskirch:
@yivi Could you squash the commits for the current pull request? That way I could test it as well and let you know if any of my use cases fail.

yivi:
@jtsternberg , #446 is the current one.
I'll update it to trunk this afternoon.

jtsternberg:
This should be good in [trunk](https://github.com/WebDevStudios/CMB2/tree/trunk). Huge thanks for @yivi's work. Please test.

Issue #346, opened 2015-06-03
Field value is wiped on post save if Field attributes are disabled => disabled


kynetiv:
I have a field with the attributes set to readonly and disabled:

``` php
$allow_ad_sync->add_field( array(
                'name' => __( 'Last Modified', 'cmb2' ),
                'desc' => __( 'Time last modified in LDAP', 'cmb2' ),
                'id'   => $prefix . 'last_modified',
                'type' => 'text_date_timestamp',
                'date_format' => 'm-d-Y @ h:i:s',
                'on_front' => false, // Optionally designate a field to wp-admin only
                'show_on_cb' => array( $this, 'upd_cbm2_is_user_admin' ), // function should return a bool value,
                'attributes'  => array(
                    'readonly' => 'readonly',
                    'disabled' => 'disabled',
                ),
            ) );
```

I populate the field value through a separate script and just wish to display it to users without giving them the ability to update it ( readonly and disabled ).

This issue here is that because it isn't passed along in the $_POST data ( I believe because it is marked as disabled in the DOM ) the value is set to null in CMB2_Field.php on line 429:
https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_Field.php#L429

``` php
/**
     * Process $_POST data to save this field's value
     * @since  2.0.3
     * @param  array $data_to_save $_POST data to check
     * @return bool                Result of save
     */
    public function save_field_from_data( $data_to_save ) {

        $meta_value = isset( $data_to_save[ $this->id( true ) ] )
            ? $data_to_save[ $this->id( true ) ]
            : null;

        return $this->save_field( $meta_value );
    }
```

Setting it to null is then set as the value and wipes my field's value. Instead there probably should be a check for disabled fields and not apply any update. The benefit of adding a conditional here would disallow tampering with the DOM (although there are probably other checks for that already).

As a workaround I've added a hook to the override filter here:
https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_Field.php#L290
 for my specific field_id on the filter (I may do one more generic for disabled fields in general, but right now I only have the one field): 

``` php
add_filter( 'cmb2_override_{_my_prefix_and_field_id}_meta_remove', array($this, 'upd_dont_remove_disabled_field'), 10, 4 );

    public function upd_dont_remove_disabled_field($override, $args, $field_args, $field_object) {
        $override = true;
        return $override;
    }
```

[...]

PR #500, opened 2015-10-20
Add new function eligible_to_save

jamesgol:
Creates a check to see if a field should be saved based on the attributes set for it.  Otherwise fields set to disabled and/or readonly will be written to.

Setting a field to readonly just sets the HTML input form to readonly, it's trivial for someone to pass a different value in and change it.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -500,8 +500,31 @@ class CMB2 {
 		$this->updated = array();
 
 		foreach ( $this->prop( 'fields' ) as $field_args ) {
-			$this->process_field( $field_args );
+			if ( $this->eligible_to_save( $field_args ) ) {
+				$this->process_field( $field_args );
+			}
 		}
+	}
+
+	/**
+	 * Determine if field is eligible to be saved
+	 * @since  2.1.3
+	 * @param  array  $field_args Array of field arguments
+	 */
+	private function eligible_to_save( $field_args ) {
+		$eligible = true;
+
+		if ( isset( $field_args[ 'attributes' ] ) ) {
+			$attributes = $field_args[ 'attributes' ];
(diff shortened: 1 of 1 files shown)


jamesgol:
Related to #346

jrfnl:
I would advise against this change as disabled and/or readonly field attributes are often the initial state of a field, with the attribute(s) being removed via javascript if certain conditions apply.

jtsternberg:
Sorry @jamesgol, I'm with @jrfnl in that I don't think it's safe to make those assumptions as readonly and disabled attributes are used for specific and varied purposes.

jtsternberg:
That being said, I could see using a field attribute, `'save_field'`, (which mirrors the [CMB2 property](https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2.php#L78)) which, when set to `false`, would keep the field from saving.

jamesgol:
Perhaps adding a filter that is checked instead would be a reasonable option?  It's been a bit since I worked on the project needing this and I had to change my process completely because there wasn't any way to correct this flaw.

jtsternberg:
There are [several](https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_Field.php#L313) [filters](https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_Field.php#L324) which could be leveraged to accomplish this already.

jamesgol:
It has been so long I don't remember the specifics on the use case there, but those filters all happened much to late.  It was related to a security issue where the original coder assumed that marking a field as read-only was really read-only and not just set in HTML.

Adding an additional setting is a reasonable solution, if nobody else gets to it first I'll submit a PR for it.

jtsternberg:
@jamesgol oops, I just pushed an update to include this, but please submit your PR and I will merge it in for the unit tests, etc, and will give you props.

Issue #496, opened 2015-10-16
Multi-input ("Address") type fields iterator not working correctly when set as repeatable.


Steviant:
Something appears to go wrong with the iterator when adding repeating "address"-type fields from https://github.com/WebDevStudios/CMB2-Snippet-Library.

It's fine with the iterations `0`, `1` and `2`, and then it sticks on `2` for all subsequent additions, so you can add 6 iterations of the field and all after the second will be assigned as `data-iterator="2"` in the markup, and the array key will be assigned at `2`, causing sanitization to fail - I think this is what's causing the 'phantom fields' issues described in https://github.com/WebDevStudios/CMB2-Snippet-Library/issues/17 and https://github.com/WebDevStudios/CMB2/issues/373 but I can't figure out where the failure is. 

Anyone else coming across this? I've tested it both with my own multi-input fields and with the 'address' field taken straight from the repo on a clean WP install and it's consistent across both, so it doesn't appear to be a plugin conflict.

katojune:
I am seeing this issue occur as well with using repeating address-type field as well as my own multi-input field.

bomsn:
I have the same issue, have you come across a solution @Steviant

PR #523, opened 2015-11-13
Fixes issue #496 (repeatable multi-input field iterator failing)

Steviant:
Simply specifies ```:input``` for ```[data-iterator]``` in cmb2.js to prevent the iterator from breaking when confronted with multi-input "address"-type fields, as shown in the code snippets repo.


lucascdsilva:
Hello, I resolved that problem that way, #532 , what do you think?! sorry, I do not know very well use the github yet!

jtsternberg:
@lucascdsilva does the solution in this PR work for you?

lucascdsilva:
@jtsternberg I did not understand ! I have not found the solution to this problem

jtsternberg:
@Steviant I don't believe this accounts for non-input repeatable fields (like `<select>`, etc).

Steviant:
It should do, the [jQuery spec](https://api.jquery.com/input-selector/) describes `:input` as _Selects all input, textarea, select and button elements_. That said, I haven't tested it and the code I wrote this fix for was in a project I'm no longer involved in, so I can't check against it. I'll try to put something together later today to test against.

jtsternberg:
Ah good point, I admit, I did not see the **`:`**. I would love it if you could test this against non-input form elements (though several of the non-input field types [can't be made repeatable](https://github.com/WebDevStudios/CMB2/blob/trunk/includes/CMB2_Field.php#L589-L602)). Thank you.

PR #535, opened 2015-12-03
Fix PHP Warning on Include Function

joshuadavidnelson:
A little pre-include_once flight check.

### DIFF
--- a/includes/helper-functions.php
+++ b/includes/helper-functions.php
@@ -29,7 +29,9 @@ function cmb2_autoload_classes( $class_name ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


jtsternberg:
Why would this be needed? What scenario did you have that would cause that error to be thrown?

joshuadavidnelson:
Perhaps it's an odd one. I was update a branch of a project that was missing CMB2_Taxonomy yet it was being called somewhere else (perhaps from an update in another part of the branch) and, in debug mode, it threw an error. I saw the error, this fixed it [hulk smash]. Totally fine to ignore if you think it's overkill :bowtie:

PR #539, opened 2015-12-06
Fix Broken Unit Test on WP Trunk

JPry:
Use `esc_url_raw()` on the expected value test.

This will ensure that what we expect is always the same as
what WordPress gives us.


jtsternberg:
@JPry `esc_url_raw( 'http://' . $value )` should just be `esc_url_raw( $value )`.

JPry:
Merged in d772c276abf3dc3354e62f037625e41f73893b7e

Issue #552, opened 2015-12-21
Metabox markup/styling on term pages doesn't match default metaboxes


MickeyKay:
![image](https://cloud.githubusercontent.com/assets/3189678/11924403/57d89794-a765-11e5-948f-056912afe283.png)

When output on term pages, CMB2's metabox markup doesn't match the default WP metabox markup, and as such some of the default styling doesn't apply.

The default opening markup is:
`<table class="form-table">`

. . . to which the following CSS is applied:
```
table.form-table {
    border: 1px solid #ccc;
    padding: 4px 8px;
    border-collapse: separate;
    margin: 16px 0;
}
```

However CMB2 outputs the following markup:
`<div class="table-form">`

As such, the default styling isn't applied.

jtsternberg:
What, specifically, is the issue that you are seeing?

MickeyKay:
When working with the `cmb2-rest-api` branch, metaboxes that are output on taxonomy term pages in the admin don't match the default metabox styling on those pages, because the markup doesn't match the default metabox markup:

**CMB2 Metabox**
(No border, padding, title is different size, etc)
![image](https://cloud.githubusercontent.com/assets/3189678/12128003/77873356-b3af-11e5-910f-9c9b909743e2.png)

**Default metabox**
![image](https://cloud.githubusercontent.com/assets/3189678/12128016/87f5a330-b3af-11e5-89d5-9bdcd041bf15.png)

jtsternberg:
The form markup for CMB2 a) intentionally avoids tables for markup, and b) intentionally matches the default WordPress fields for continuity: 

![](http://b.ustin.co/6Cbm+)

PR #554, opened 2015-12-21
Fix metabox form output for taxonomy terms

MickeyKay:
Addresses #552. Wraps the existing taxonomy term metabox output in a
table and associated tbody, tr, and td elements to match default
WordPress markup.


MickeyKay:
Accidental changes, going to try again.

PR #557, opened 2015-12-22
Fix issue with empty '' and zero '0' values in select and radio

IgorCode:
Issue was with both empty string '' and string with zero '0' being treated equally so both would always be selected.
This issue was present in following configuration:
```
array(
	'type'    => 'select',
	'options' => array(
		''  => 'Default',
		'1' => 'Yes',
		'0' => 'No',
	)
)
```

In proposed fix I changed it so that escaped_value is checked for an actual empty string, this way default argument is used only when there is no data in meta field and not when data is '0'. And values are converted to strings before comparison so that '0' can't be equal to '', but '0' will still match integer 0 in case someone puts integer values in their config.

### DIFF
--- a/includes/CMB2_Types.php
+++ b/includes/CMB2_Types.php
@@ -223,7 +223,7 @@ class CMB2_Types {
 			$a['label'] = $opt_label;
 
 			// Check if this option is the value of the input
-			if ( (string) $value === (string) $opt_value ) {
+			if ( !is_array( $value ) && (string) $value === (string) $opt_value ) {
 				$a['checked'] = 'checked';
 			}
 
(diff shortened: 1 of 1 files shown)


tw2113:
I feel like this one could be better handled just by not using empty strings and string '0' for the option names. Save many headaches with that one.

jtsternberg:
This should be fixed in trunk.

tw2113:
Closing then. If we find it's not, we can start a new ticket regarding it.

IgorCode:
@tw2113
It sure is easier to avoid empty values, but I don't see how can it be better? :/

Anyway, this code has been in production environment for almost a year. I'm currently on some non WordPress project so I don't have time to move to trunk (which, I believe, didn't exist back when I made this commit) and re-test again, if anyone wants to try to move to trunk by salvaging some of my code it's fine by me :)

PR #577, opened 2016-01-31
Allow for optional/empty money field.

jrfnl:
The number formatting of fields of the text_money type caused a money field to *always* be saved.
If the field was empty, it would be saved as '0.00'.

This fix allows for the proper (non-)saving of empty money fields.

### DIFF
--- a/includes/CMB2_Sanitize.php
+++ b/includes/CMB2_Sanitize.php
@@ -169,6 +169,9 @@ class CMB2_Sanitize {
 	 * @return string       Empty string or sanitized money value
 	 */
 	public function text_money() {
+		if ( '' === $this->value ) {
+			return $this->value;
+		}
 
 		global $wp_locale;
 
@@ -178,6 +181,9 @@ class CMB2_Sanitize {
 		// for repeatable
 		if ( is_array( $this->value ) ) {
 			foreach ( $this->value as $key => $val ) {
+				if ( '' === $val ) {
+					continue;
+				}
 				$this->value[ $key ] = number_format_i18n( (float) str_ireplace( $search, $replace, $val ), 2 );
 			}
 		} else {


monecchi:
This issue seems not to be solved yet for the cmb2-rest-api branch, I'm still getting money field being saved even if it is empty causing it to have a '0.00' value.

jrfnl:
Did you test with this branch or with the trunk / master  branch ? This PR (unfortunately) has not been merged in yet.

jtsternberg:
Just saw this. Thanks!

monecchi:
Hey, @jrfnl. I did actually tested with this branch, which by the way seems not to expose CMB2 custom fields to the WP Rest API either. Maybe I am misunderstanding the way the CMB2 branches work. I've forked the cmb2-rest-api branch and updated my custom post type and cmb2 fields in order to expose it to the WP Rest API v2, as long as I thought it would have all the goods from the master branch + the WP API support.

jrfnl:
I've not worked with the `cmb2-rest-api` branch myself yet. Maybe @jtsternberg would be the better person to ask in this case.

The only thing I can say about it, is that by the looks of it the `rest-api` branch is a little behind the `trunk` branch (which now has is fix) and trunk would need to be merged in for the `rest-api` branch to also benefit from this fix.

monecchi:
@jrfnl oh, I got it, thanks for the input. So I think it's a matter of waiting for those branches to get merged. 

The fact is the cmb2-rest-api branch is not working as it should yet, it does not expose my fields to the WP API, so I'm thinking about going back to the master or the trunk branch for the meantime.

If at least I could get the API functionality at my disposal, then the saved '0.00' value for the empty text_money field is a thing I could live with.

jtsternberg:
The cmb2-rest-api branch is still in-progress, and as such, documentation is sparse. The fields showing in the REST API is an opt-in feature, not opt-out, so you wouldn't see anything by default. Just search for `'show_in_rest'` in the [example-functions.php file of that branch](https://github.com/WebDevStudios/CMB2/blob/cmb2-rest-api/example-functions.php). Also, the rest api branch is up-to-date with trunk now.

monecchi:
Hey @Jsternberg. Thanks for pointing out the file, I've already followed all the necessary steps to make it work but no success. I'll try the trunk branch to see if it keeps up with all my cmb2 plugins while solves the WP Rest API not exposing the fields issue.

jsternberg:
@jtsternberg I think the above message was meant for you.

PR #578, opened 2016-01-31
Add validation for radio, select and select_timezone fields.

jrfnl:
Validated the received value against the allowed options as provided in the field definition.
Returns false if the received value is invalid.

### DIFF
--- a/includes/CMB2_Sanitize.php
+++ b/includes/CMB2_Sanitize.php
@@ -71,6 +71,11 @@ class CMB2_Sanitize {
 			case 'oembed':
 				$sanitized_value = $this->textarea();
 				break;
+			case 'radio':
+			case 'radio_inline':
+			case 'select':
+				$sanitized_value = $this->validate_against_options();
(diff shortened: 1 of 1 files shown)


jtsternberg:
There are definitely some things to like about this concept, but as it is now, it could break backwards compatibility, and it does not account for dynamically built/modified options arrays which would result from using the options_cb paramater.

jrfnl:
> There are definitely some things to like about this concept, but as it is now, it could break backwards compatibility.

Could you give me some example of why this would break backward compatibility ? For radio and select fields, the options are 'fixed', so any option not in the options list would potentially be a XSS.

All I can think of is options dynamically added or removed via js which I believe would be rare. All the same, I've added a filter now to allow modifying the options array as used for validation.

I'll gladly try and work round any other situations you can think of.

> it does not account for dynamically built/modified options arrays which would result from using the options_cb paramater. 

Good point. Fixed.

PR #582, opened 2016-02-04
Add a filter to allow for adding arbitrary HTML attributes to a group…

jrfnl:
… wrapper element.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -352,9 +352,24 @@ class CMB2 {
 		$remove_disabled = $nrows <= 1 ? 'disabled="disabled" ' : '';
 		$field_group->index = 0;
 
+		/**
+		 * Allow for adding additional HTML attributes to a group wrapper.
+		 *
+		 * If you use this filter, make sure the resulting string always starts with a space!
+		 *
+		 * @since 2.2.2
+		 *
+		 * @param string Current attributes string.
+		 *
+		 * @param array $group_args All group arguments.
+		 *
+		 * @param string $id The group_id of the current group.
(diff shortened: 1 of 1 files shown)


tw2113:
@jrfnl may not be a bad idea to put a space before the `$more_attributes` variable to prevent running attributes together. Not quite sure how browsers would handle it, but perhaps better safe than sorry? We should also add attribute escaping for security purposes.

I am curious about the use-cases that you see with this.

@jtsternberg Since this one defaults to an empty string, and is just html attributes, i don't see a huge amount of breakage potential.

jtsternberg:
I'm inclined to move the `CMB2_Types::concat_attrs()` method to `CMB2_Utils`, and use that generate the string of attributes, and have the filter take an array. BUT..  this filter feels a bit arbitrary to me. Is there any reason to do this instead of doing things in JS, which is likely where you'll be wanting to use the extra attributes anyway?

jrfnl:
> Is there any reason to do this instead of doing things in JS, which is likely where you'll be wanting to use the extra attributes anyway?

@jtsternberg This is for the CMB2 Conditionals plugin to allow for conditional groups. The attributes determining the conditions would need to be added for the js to be able to pick up on it.

jtsternberg:
Thanks @jrfnl! Hopefully that helps w/ the CMB2 Conditionals plugin. I've updated the filter to use `concat_attrs`, which means the filter sends an array and requires one back. See the change here: https://github.com/WebDevStudios/CMB2/commit/337c24317f8de88a456f9f3691ace635a2f8dcc5

PR #584, opened 2016-02-06
Use WP_Screen for script/styles loading.

jrfnl:
I might be missing something, but it looked like the scripts/styles were not being loaded on all possible back-end pages which could have CMB2 metaboxes.
Also, the previous array contained two invalid values.

Suggestion: better yet would be to use the `WP_Screen` object via `get_current_screen()` to check for the right pages.

### DIFF
--- a/includes/CMB2_hookup.php
+++ b/includes/CMB2_hookup.php
@@ -200,9 +200,16 @@ class CMB2_hookup {
 	 * Enqueues scripts and styles for CMB2
 	 * @since  1.0.0
 	 */
-	public function do_scripts( $hook ) {
-		// only enqueue our scripts/styles on the proper pages
-		if ( in_array( $hook, array( 'post.php', 'post-new.php', 'page-new.php', 'page.php', 'comment.php' ), true ) ) {
+	public function do_scripts() {
+		$screen = get_current_screen();
+
+		if ( ! property_exists( $screen, 'base' ) || ! property_exists( $screen, 'post_type' ) ) {
+			return;
+		}
+
+		// Only enqueue our scripts/styles on the proper pages.
+		if ( ( 'post' === $screen->base && ( is_string( $screen->post_type ) && '' !== $screen->post_type ) ) ||
+			'comment' === $screen->base ) {
 			if ( $this->cmb->prop( 'cmb_styles' ) ) {
 				self::enqueue_cmb_css();
 			}


jtsternberg:
Agree about using `WP_Screen` object. But, the enqueueing for the term/user pages is handled here: https://github.com/WebDevStudios/CMB2/blob/master/includes/CMB2_hookup.php#L260-L265

In my testing, the js/css is loaded in all the expected places, term pages, user profile, comment pages, post pages, etc.

jrfnl:
Ah, there we go. It did feel very much like I was missing something, so good.

Let me just change over this PR to using WP_Screen in that case, or would you rather receive a new PR for that ?

jtsternberg:
Let's do a new PR to keep the commit history a bit cleaner (unless you feel like scrubbing your branch history).

jrfnl:
I'll just rebase & squash, no worries.

jrfnl:
So this should only check for the edit comments and post pages ? Not term edit or user pages. Correct ?

jtsternberg:
Yah, basically the two contexts where `add_meta_box` is used, posts and comments. Need to be sure it works on new and edit pages.

jrfnl:
No worries, in WP_Screen they are not very different, new items just have the `add` `action`.

PR refreshed.

jtsternberg:
`parent_base

jrfnl:
@jtsternberg That's weird, it shouldn't be.
Hmm.. just looked into it deeper. Looks like WP does not have all information for WP_Screen setup on the `admin_enqueue_scripts` hook, which is weird as the `current_screen` action has already run by that time.

Let me play round with this a bit more to see which other info is available and usable for this situation.

jrfnl:
Another weird thing I'm running into - looks like the `do_scripts` function - in my test set up - is actually being called 8 times, even though it looks like you've tried to prevent that by having the `once` method. So I guess there's a bug in the `once` method or is this not what you were trying to accomplish there ?

If you like, I can fix that too - just a simple matter of wrapping the `add_filter` within `once` within a `has_filter` call.

jtsternberg:
Yah, that's odd. I think it probably has something to do w/ the serializing of the func args and the `$this` in the callback being unique. Anyway, I think it would be good to fix the `once` issue.. possibly w/ something like a `has_filter` proxy method. I'd be happy to see what you come up with.

jrfnl:
Ok, will do that in a separate PR.

For now, I've refreshed this PR with some code which *should* work on the `admin_enqueue_script` hook.
The scripts will load on add/edit pages, but not on the overview pages with the current code.

jrfnl:
> I think it probably has something to do w/ the serializing of the func args and the $this in the callback being unique. Anyway, I think it would be good to fix the once issue.. possibly w/ something like a has_filter proxy method.

You're right, it does have to do with the $this being different objects.
The oembed handler actions are fine as they use the same object via the `cmb2_ajax()` function and the `register_scripts` method is fine too with it being a static method.

It's just the calls from `post_hooks()` and `comment_hooks()` which pass the object which are troublesome.

tw2113:
What's the status of this one @jrfnl ? Seeing any issues with it in your own dealings?

jrfnl:
Sorry haven't had a chance to work on this lately. Will get back to it at some point though ;-)

PR #588, opened 2016-02-11
Introduce three new filters to filter field arguments.

jrfnl:
The `cmb2_field_defaults` filter to filter the generic field defaults. Especially useful if you have a lot of fields which use nearly the same arguments. This way you can just set the defaults for those fields in one go.

The `cmb2_field_arguments_raw` filter to filter the arguments used to set up the field *before* processing.

The `cmb2_field_arguments` filter to filter the arguments used to set up the field *after* processing.

All filters also pass the field id, so the filter can (and should) selectively be applied by testing the id against a specific id or testing it against the prefix substring.
They also both pass the field type so the filter can selectively be applied against one specific field type only.

Example use case for the defaults filter:
* Change the `show_names` variable default to `false` for all CMB2 fields.
* If you have a default set of radio options for all radio fields in a specific metabox, you could set this in one go for all your radio fields.
```php
add_filter( `cmb2_field_defaults`, 'my_field_defaults_filter', 10, 3 );
function my_field_defaults_filter( $args, $id, $type ) {
	if ( strpos( $id, 'myprefix' ) === false || $type !== 'radio' ) {
		return $args;
	}

	$args['options'] = 'my_radio_options_callback';
	return $args;
}
```

Example use case for the field arguments raw filter:
* Changing all textarea field to textarea_small:
```php
add_filter( `cmb2_field_arguments_raw`, 'my_field_arg_raw_filter', 10, 3 );
function my_field_arg_raw_filter( $args, $id, $type ) {
	if ( strpos( $id, 'myprefix' ) === false || $type !== 'textarea' ) {
		return $args;
	}

	$args['type'] = 'textarea_small';
	return $args;
}
```

Example use case for the field arguments filter:
* Say you've added another field type which needs supporting data, this filter would allow you to set `$args['has_supporting_data']` to true.


jtsternberg:
IMO, this processing should happen during the field registration process. For your radio example:

```php
	$fields[] = array(
		'name'             => __( 'Test Radio inline', 'cmb2' ),
		'desc'             => __( 'field description (optional)', 'cmb2' ),
		'id'               => $prefix . 'radio_inline',
		'type'             => 'radio_inline',
		'show_option_none' => 'No Selection',
		'options'          => array(
			'standard' => __( 'Option One', 'cmb2' ),
			'custom'   => __( 'Option Two', 'cmb2' ),
			'none'     => __( 'Option Three', 'cmb2' ),
		),
	) );

	$fields[] = array(
		'name'    => __( 'Test Radio', 'cmb2' ),
		'desc'    => __( 'field description (optional)', 'cmb2' ),
		'id'      => $prefix . 'radio',
		'type'    => 'radio',
		'options' => array(
			'option1' => __( 'Option One', 'cmb2' ),
			'option2' => __( 'Option Two', 'cmb2' ),
			'option3' => __( 'Option Three', 'cmb2' ),
		),
	) );

	foreach ( $fields as $field ) {
		switch ( $field['type'] ) {
			case 'radio':
				$field['options'] = 'my_radio_options_callback';
				break;
		}

		$cmb_demo->add_field( $field );
	}
```
The only filter that makes sense to me is the last filter, cmb2_field_arguments, but I think it should pass `$this` as the 2nd argument to allow full access to the field object.

jrfnl:
> IMO, this processing should happen during the field registration process.

@jtsternberg I know how it can be done during registration, that's not the point.

These filters are useful for CMB2-add-on plugins as well as making large changes quickly.

Say, you've got ten metaboxes set up with some 150 fields. Instead of having to edit all the fields to make a change to a particular field type, you could now just filter them.

I've currently got my setup organized so it will filter all arguments from all metaboxes before passing them to CMB2, but it makes a lot more sense to have such a filter in CMB2 instead of in each implementation.

jrfnl:
Rebased, added filter documentation and changed the last argument as per suggestion.

jtsternberg:
Let's yank the `cmb2_field_arguments_raw` filter, and we'll call it good. I can't imagine a scenario where it would be needed over the `cmb2_field_arguments` filter.

jrfnl:
If you insist, I'll take it out, but the reason I put it in was to be able to globally change some field arguments without having to duplicate the logic contained in the rest of the function.
If one changes it using the `cmb_field_arguments` filter at the end, the logic within the function would often still need to be duplicated.

PR #602, opened 2016-03-01
Implement CMB2_Ajax as a singleton.

jrfnl:
CMB2_Ajax is already being used as a singleton, but doesn't enforce it which leads to the static `hooks()` function, so let's enforce it and be done with it.


jtsternberg:
Fair enough. Though I'd make one change: `cmb2_ajax()` should simply `return CMB2_Ajax::get_instance();`.

jrfnl:
Ok, done.

jtsternberg:
Ok, thanks!

PR #606, opened 2016-03-13
Added support for select multiple

jamarzka:
- Updated concat_items to support an array of values when checking if the option is checked.
- Added a new multiselect type that sets the multiple attribute and appends [] to the name.


tw2113:
Are you familiar at all with unit testing and writing them? Having some for this would definitely help get it merged in.

We'd also need to get the conflicts resolved, of course

PR #610, opened 2016-03-19
Autoload init file

lonnylot:
When using composer we should autoload the init file so we have immediate access to the CMB2 functionality

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,5 +25,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.*"
+    },
+    "autoload": {
+        "files": ["init.php"]
     }
 }


jtsternberg:
See #198, https://github.com/WebDevStudios/CMB2/commit/e863b18ca838f2e5ef280e0883f597c99f44f97a, and https://github.com/WebDevStudios/CMB2/commit/e4d03897689a16d0231c2a344fc15b0439e2155f.

Thank you for your contribution, but for future contributions, can you please review the [CONTRIBUTING doc](https://github.com/WebDevStudios/CMB2/blob/master/CONTRIBUTING.md) before submitting? Thank you.

PR #637, opened 2016-04-22
Update travis file for non-specific PHP minor

jrfnl:
Use unspecified PHP 7 minor in favour of specified one to ensure always using the latest available via Travis.


jtsternberg:
Thanks. I changed it because tests were inexplicably breaking. It appears they are working fine now.

PR #656, opened 2016-06-02
Make the Title field type handles extra args

vladolaru:
Currently, the Title type didn't take into consideration extra arguments like 'attributes'. Now it does, just like the rest of the fields. This will allow for add-ons like CMB2-conditionals to work for this field type also.

### DIFF
--- a/includes/types/CMB2_Type_Title.php
+++ b/includes/types/CMB2_Type_Title.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * CMB title field type
  *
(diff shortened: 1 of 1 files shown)


jtsternberg:
Nice catch. Remove the `$args` passed into the method, and I'll merge it in. It's not needed, because the `$args` [are passed to it from the `CMB2_Types` object](https://github.com/WebDevStudios/CMB2/blob/trunk/includes/CMB2_Types.php#L436-L439).

includes/types/CMB2_Type_Title.php
```diff
@@ -12,16 +13,24 @@
...
 			'tag'   => $this->field->object_type == 'post' ? 'h5' : 'h3',
 			'class' => 'cmb2-metabox-title',
 			'name'  => $this->field->args( 'name' ),
 			'desc'  => $this->_desc( true ),
-		) );
+		), $args );
 
 		return $this->rendered(
-			sprintf( '<%1$s class="%2$s">%3$s</%1$s>%4$s', $a['tag'], $a['class'], $a['name'], $a['desc'] )
+			sprintf( '<%1$s %2$s/>%3$s</%1$s>%4$s', $a['tag'], $this->concat_attrs( $a, array( 'tag', 'name', 'desc' ) ), $a['name'], $a['desc'] )
```
jtsternberg: why is there an extra forward slash before `%3$s`?

vladolaru: Because I am not careful :)

vladolaru:
I have removed the $args and the slash. Keep up the good work.

jtsternberg:
Thanks. Still a no-go. You don't need to pass the `$this->args` args as they are assumed: https://github.com/WebDevStudios/CMB2/blob/trunk/includes/types/CMB2_Type_Base.php#L86

vladolaru:
Damn. Ok. Removed it for good now.

jtsternberg:
A bit of cleanup and this is merged: https://github.com/WebDevStudios/CMB2/commit/c1718129c5cbf8e4ec6f0f3fac69b6434e3346ea

Thank you!

Issue #264, opened 2015-03-26
Make WYSIWYG / TinyMCE / Editor field repeatable?


spwebguy:
Hi there,

I know this has been asked a couple of times, I just hope you didn't give up on this. I think it's very important for CMB2. To me it's pretty much the only thing that's missing. Apart from this, being a CMB2 user and having tried many other frameworks, you da best.

Keep up the great work.
NB: I'd be willing to pay for this feature. A lot.

jtsternberg:
We haven't given up and we/I get asked very frequently. It's just a matter of getting the time. CMB2 is a (heavily used) side project for us, so we work on it when we can, and repeatable wysiwyg is at the top of the feature list roadmap.

spwebguy:
Thanks a lot for taking the time to answer!

adchsm:
Just in case this helps, I have a repeatable group with a WYSIWYG editor as a field within that group and it works perfectly.

```
function register_metabox() {
		$prefix = '_repeatable_editor_';
		
		$cmb_showcase = new_cmb2_box( array(
			'id'            => $prefix . 'metabox',
			'title'         => __( 'Repeatable Editor', 'cmb2' ),
			'object_types'  => array( 'page' ), // Post type
			'context'       => 'normal',
			'priority'      => 'high',
			'show_names'    => true, // Show field names on the left
			//'cmb_styles'    => false, // false to disable the CMB stylesheet
		) );
		
		// Repeatable group
		$group_showcase = $cmb_showcase->add_field( array(
			'id'          => $prefix . 'sections',
			'type'        => 'group',
			'options'     => array(
				'group_title'   => __( 'Editor', 'cmb2' ) . ' {#}', // {#} gets replaced by row number
				'add_button'    => __( 'Add another Editor', 'cmb2' ),
				'remove_button' => __( 'Remove Editor', 'cmb2' ),
				'sortable'      => true, // beta
			),
		) );
		
		// Main editor
		$cmb_showcase->add_group_field( $group_showcase, array(
			'name'    => __( 'Content', 'cmb2' ),
			'id'      => 'content',
			'type'    => 'wysiwyg',
			'options' => array( 'textarea_rows' => 8, ),
		) );
	}
	
	add_action( 'cmb2_init', 'register_metabox' );
```

natebeaty:
@adchsm does your new editor instantiate ok when you add it? I get a blank ckeditor area when clicking `Add another Editor`, like this:

[...]

PR #657, opened 2016-06-03
Fix for #264, repeatable groups with a WYSIWYG editor

johnsonpaul1014:
I believe this fixes the problem by setting up a new editor when the new group row is added.  It is also necessary to destroy and reinit all the editors when sorting.


alexmansfield:
It looks like custom WYSIWYG buttons (such as the one Gravity Forms adds) don't get attached after the first WYSIWYG field.

johnsonpaul1014:
That's strange.  I have some custom buttons I have added to our editor on the site I needed to implement this change for, and they are added as new editors are added to the group.  Are you using multiple editors in the same group row?

alexmansfield:
I'm not sure if it happens for buttons that are added directly to the TinyMCE editor. I was noticing it for buttons that get added above the editor (next to the "Add Media" button). For example, the Column Shortcodes plugin attaches a button there (other plugins do as well, but that was the first free example I could think of).

johnsonpaul1014:
I did some investigation into this, and it looks like the plugin you are referring to uses the edButtons JavaScript array to add the buttons.  It would probably be possible to clone this somehow for the new editors, but I think that should be left to someone else as I don't need that feature for my current project.  The change is still a huge step forward, because at least you get a real editor and not a broken one when you add a group row.

alexmansfield:
Right, I wasn't trying to put down your improvements at all. Your patch is a huge step forward. I just noticed the missing buttons when testing it out and figured this was probably the place to mention it. I'm sorry if this was the wrong place to mention it. Thanks for the effort you put into this, and keep up the good work!

johnsonpaul1014:
Thanks.  I agree; this is definitely the place to mention it.  Hopefully someone else can grab the baton from here.

webakimbo:
@johnsonpaul1014 would love to use your branch for a time-sensitive project but the build appears to be broken. Is the plugin still working on your end?

johnsonpaul1014:
I forked off master instead of trunk, so that was also an issue.  I am going to make a new pull request off trunk to see if that works this morning.

Paul

Date: Sun, 17 Jul 2016 19:17:33 -0700

johnsonpaul1014:
I have submitted the new pull request.

Paul

Date: Sun, 17 Jul 2016 19:17:33 -0700

webakimbo:
Thanks! Will keep an eye out for activity.

tw2113:
@johnsonpaul1014 are we good to close this one, if you've issued a fresh pull request with updated details?

johnsonpaul1014:
Yes.

PR #658, opened 2016-06-03
File-Upload field should attach new uploads to current post

vladolaru:
I think this is a more consistent way of working with uploads.

Thanks


jtsternberg:
It's not clear to me what you are trying to accomplish with this PR, but I'm not in favor as using a post frame adds a bunch of irrelevant UI to the modal.

vladolaru:
You are right about those, but I think we could find a way to "deactivate" those. What I am after is to attach the uploaded file to the current post. Will investigate further.

jtsternberg:
Using CMB2 as-is, when I upload a new file via the file modal, it automatically associates that post with the post. I'm not sure how yours differs.

vladolaru:
The fact that it is working for your is just a side effect fo the fact that you are using them on a post type that supports the 'editor'. I am using it on a CPT with just title. I think it would be best if things could be a little more encapsulated.

vladolaru:
But my solution didn't solve everything. I've ended up extending the Select MediaFrame to it to use a single state with a certain set of parameters. Would you be interested in using such an approach or should I make it as separate code for my own project?

jtsternberg:
Sure, I would be interested. As long as the UI is similar (or better) for our use-case.

vladolaru:
Ok. I am also very very keen on keeping the UI as clean as possible. That
is why I am going to such lengths to improve things but without loosing
ease of use.

jtsternberg:
Going to close. Feel free to open a new PR if you are able to make those changes.

PR #672, opened 2016-06-28
Small string improvement

pedro-mendonca:
Move the endpoint into the string for coherence, and if translators won't check the code, might add a double endpoint by mistake.


jtsternberg:
Thank you!

PR #675, opened 2016-06-28
Field comments

jamesgol:


jtsternberg:
Hmm, why did you add `array|int|` to the return comment? `$updated` will only ever be a boolean, unless I'm missing something.

jamesgol:
When I was debugging it returns the value of update_metadata which can be false, the meta ID of the object, and in some weird case it was actually an array.  I'll have to trace back when I have a chance to see why it would be an array

jtsternberg:
@jamesgol nah, no need, I didn't dig that deep in to `update_metadata` to verify, but sounds like you're on it. Thanks!

jamesgol:
The 'test_get_oembed_delete_with_expired_ttl' test gets a return of an array.  Looks like the remove_data() function is returning an array, that doc probably needs updated too

jtsternberg:
I'll take any doc updates you want to send. :)

PR #683, opened 2016-07-06
Customization of group title hash replacement

brandondove:
Allow population of group title bar via sub fields

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -916,14 +916,23 @@ class CMB2_Field extends CMB2_Base {
 	}
 
 	/**
-	 * Replaces a hash key - {#} - with the repeatable index
+	 * Replaces a hash key - {#} - with the repeatable index, but allow customization of replacement via filter
 	 * @since  1.2.0
 	 * @param  string $value Value to update
 	 * @return string        Updated value
 	 */
 	public function replace_hash( $value ) {
-		// Replace hash with 1 based count
-		return str_ireplace( '{#}', ( $this->index + 1 ), $value );
+		// Replace hash with 1 based count by default, but allow customization of replacement via filter
+		return str_ireplace(
+			'{#}',
+			apply_filters(
+				'cmb2_group_title_hash_replacement',
+				( $this->index + 1 ),
+				$this,
+				$this->index
+			),
+			$value
+		);
 	}
 
 	/**


jtsternberg:
The problem is that doesn't cover the JS side of things, so when you add/remove/sort the groups, that's not going to work out so well. I'd instead suggest doing something like this: https://github.com/WebDevStudios/CMB2-Snippet-Library/blob/master/javascript/group-field-id-match-input-value.php

brandondove:
Actually, I think sorting works out better if you don't use numbered groups because it doesn't matter what order they're in.

In the example below, new groups added with javascript would get "unsaved" in the title. Once you save, the meta box title gets the proper title. This would allow better visibility to what's inside of a closed group meta box instead of having to open and close 50 numbered boxes to find what you're seeking.
```
add_filter( 'cmb2_group_title_hash_replacement', 'add_subfield_to_title', 10, 3 );

function add_subfield_to_title( $default, $field, $index ) {
	if ( $field->args['id'] == 'example-group' ) {
		if ( ! empty( $field->value[$index]['title'] ) ) {
			return '(' . esc_html( $field->value[$index]['title'] ) . ')';
		} else {
			return '(unsaved)';
		}
	}
	return $default;
}
```

jtsternberg:
Totally, I definitely see the benefit. My argument was that doing it in PHP-only was kind of a non-starter for me, as it requires a save before anything happens. BUT, i'll check it out w/ your provided example.

jtsternberg:
Hey, just played with this a bit. The replaced titles still 'stick' when the groups are sorted. I think the best bet is to get [something like this in core](https://github.com/WebDevStudios/CMB2-Snippet-Library/blob/master/javascript/group-field-id-match-input-value.php).

BraadMartin:
I just ran into this same issue today. I understand your concerns about the group titles only updating after save @jtsternberg, but this fits my use case just fine, and I believe it's as simple as either adding the filter that @brandondove added here or adding a set method on the `CMB2_Field` class that would allow us to affect the `options` property from https://github.com/WebDevStudios/CMB2/blob/65f4831a08df1e4ded0c786f0c377b47aface75f/includes/CMB2.php#L407

I realize that a JS solution would be the ultimate, but can you think of any harm in providing a way to do it with just PHP also? This would be really helpful for my use case.

Thanks for your excellent work on CMB2 btw! I use it on every project, and it has become an essential tool for me and the company I work for.

jtsternberg:
Hey @BraadMartin, I'm not trying to be a stick in the mud, but I don't see a purpose in adding a filter which is half-broken out of the box. It's nearly ok, but the fact that the titles stay in place when the groups are started make this a no-go for me.



If you're looking for a working solution, I linked to one above.

brandondove:
@jtsternberg you're right, this definitely needs more work. I went ahead and implemented the javascript you recommended and it works perfectly.

BraadMartin:
@jtsternberg Thanks for the reply, and I totally understand. I'll use the JS solution you suggested. :)

jtsternberg:
Closing for now, but #706 is open for the feature request.

PR #693, opened 2016-07-19
Fix for #264, repeatable groups with a WYSIWYG editor, take 2

johnsonpaul1014:
This is a second attempt at a pull request to fix the repeatable WYSIWYG editor fields.  It adds in a fix to make sure that editors aren't attempted to be destroyed before they actually exist.  I have been using this in production for a bit now with no more issues.  There may be a better way to integrate this with existing CMB code, but I was able to pull this out of some other logic I had to make WYSIWYG editors.


matgargano:
:+1: getting this in would be awesome

jtsternberg:
Not ignoring, just busy. We'll be reviewing soon.

jtsternberg:
Hmm, you don't actually ever enqueue the wysiwyg.js. Is there a reason for that?

jtsternberg:
I see, you've concatenated it, but their is no solution if `SCRIPT_DEBUG` is enabled. I'm reviewing now, so please don't make any updates.

johnsonpaul1014:
That's a good point.  It probably makes more sense to enqueue it with the main script having a dependency on it, so a local wysiwyg variable could be used instead of window.wysiwyg.  I was just trying to eliminate the extra server request.

jtsternberg:
[This](https://github.com/WebDevStudios/CMB2/pull/698) is now in trunk and will be in the next release.

PR #696, opened 2016-07-29
i18n Improvements

ramiy:
update translation strings related issues.

- Security: escaping translation strings with `esc_html__()` and `esc_attr__()`.
- Replace `<a>` links with `%s` placeholder.
- Merge three similar translation strings into one string, and use placeholders for the links
- Add translators comments to string with placeholders (`%s`, `%1$s` and `%2$s`).


example-functions.php
```diff
@@ -268,144 +268,148 @@ function yourprefix_register_demo_metabox() {
...
 		'preview_size' => array( 100, 100 ), // Default: array( 50, 50 )
 	) );
 
 	$cmb_demo->add_field( array(
-		'name' => __( 'oEmbed', 'cmb2' ),
-		'desc' => __( 'Enter a youtube, twitter, or instagram URL. Supports services listed at <a href="http://codex.wordpress.org/Embeds">http://codex.wordpress.org/Embeds</a>.', 'cmb2' ),
+		'name' => esc_html__( 'oEmbed', 'cmb2' ),
+		'desc' => sprintf(
+			/* translators: %s: link to codex.wordpress.org/Embeds */
+			esc_html__( 'Enter a youtube, twitter, or instagram URL. Supports services listed at %s.', 'cmb2' ),
```
jtsternberg: When you want to leave comments for translators, the `_x()` family of functions should be used. e.g.
```php
esc_html_x( 'Enter a youtube, twitter, or instagram URL. Supports services listed at %s.', '%s is a link to codex.wordpress.org/Embeds', 'cmb2' ),

```

includes/CMB2_Ajax.php
```diff
@@ -170,11 +170,19 @@ public function get_oembed( $args ) {
...
 		}
 
 		// Otherwise, send back error info that no oEmbeds were found
-		return '<p class="ui-state-error-text">' . sprintf( __( 'No oEmbed Results Found for %s. View more info at', 'cmb2' ), $oembed['fallback'] ) . ' <a href="http://codex.wordpress.org/Embeds" target="_blank">codex.wordpress.org/Embeds</a>.</p>';
+		return sprintf(
+			'<p class="ui-state-error-text">%s</p>',
+			sprintf(
+				/* translators: 1: results for. 2: link to codex.wordpress.org/Embeds */
+				esc_html__( 'No oEmbed Results Found for %1$s. View more info at %2$s.', 'cmb2' ),
+				$oembed['fallback'],
```
jtsternberg: As mentioned above, please use `esc_html_x()` here.

ramiy:
@jtsternberg 
Actually, you are wrong.

You should never use context `_x()` function to describe placeholders. You should use only translator comments `/* translators: %s: placeholder description */`. The translator comment will appear on the sidebar:

context

As for the `_x()`, in WordPress core, we use context functions when a string is used in many places, but we want one of those strings to have a different meaning. When we add the context function, this string will appear as a separate string in translate.wordpress.org.

context

When using context function, It creates a new line in translate.wordpress.org, and allows you to provide a different meaning for this particular string.

jtsternberg:
Oh wow, well thank you very much for educating me! Looks like I've been using those functions improperly. I'll review this a bit more thoroughly when I get back to my computer, but should be good. Thanks again!

jazzsequence:
@jtsternberg the best example of proper use of `_x` would be:

```
_x( 'Buffalo', 'textdomain', 'A city in New York.' );
_x( 'Buffalo', 'textdomain', 'A large animal that lives in North America' );
```

jtsternberg:
Thanks for you contribution @ramiy!

PR #722, opened 2016-08-25
[#326] Included the usage of the before_field for the default currency

### LINKED ISSUE #326: Possibility of setting a default currency symbols for text_money
I have realized that as a developer you can not manage to set a default currency_symbol for the text_money field, it is always *$*. The easiest (but weird in a way) is to make a loop and add the argument 'before_field' but there is a difficult in text_money fields inside 'group' fields.

I think we can solve it with the addition of the the code below inside the function _set_field_defaults.

```
if( $args['type'] === 'text_money' ) {
	$args['before_field'] = apply_filters( 'cmb2_text_money_before_field_filter', ( isset($args['before_field']) ? $args['before_field'] : '$' ), $this );
}
```

*Note:*

I thought on adding the feature of detecting the currency based on wp_locale but it is not the best way when it comes to multilanguage sites.

By the way, one can get the currency according to the location with the functions:

```
if( !defined( 'CMB2_NUMBER_LOCALE' ) ) {
	define( 'CMB2_NUMBER_LOCALE', get_locale() );
}

[...]

private function get_text_money_before_field() {
	if(!class_exists('NumberFormatter')){
		return '$';
	}

	$numberFormatter = new NumberFormatter( CMB2_NUMBER_LOCALE, NumberFormatter::CURRENCY );

	return $numberFormatter->getSymbol( NumberFormatter::CURRENCY_SYMBOL );
}
```
I think that the currency detection is out of the scope of this plugin so it must be setted somewhere else..

jcchavezs:
Fixes #326 .
-  Possibility of defining a currency symbol for the `text_money` field type.

Ping @jtsternberg

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -1131,6 +1131,10 @@ class CMB2_Field extends CMB2_Base {
 			$args['options']['textarea_name'] = $args['_name'];
 		}
 
+		if( 'text_money' === $args['type'] ) {
+ 			$args['before_field'] = apply_filters( 'cmb2_text_money_before_field_filter', ( isset($args['before_field']) ? $args['before_field'] : '$' ), $this );
+ 		}
+
 		$option_types = apply_filters( 'cmb2_all_or_nothing_types', array( 'select', 'radio', 'radio_inline', 'taxonomy_select', 'taxonomy_radio', 'taxonomy_radio_inline' ), $this );
 
 		if ( in_array( $args['type'], $option_types, true ) ) {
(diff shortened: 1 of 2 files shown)


jtsternberg:
The only way I'd be willing to merge this in is if the [back-compatibility concerns](https://github.com/WebDevStudios/CMB2/issues/326#issuecomment-126105079) are addressed. The only way I see that happening is to use a variable filter name with the `cmb_id` being part of the hook (example: https://github.com/WebDevStudios/CMB2/blob/trunk/includes/CMB2.php#L140). And `cmb_id` is not guaranteed in this specific method context, so this would likely be better off happening in the `add_field` method: https://github.com/WebDevStudios/CMB2/blob/trunk/includes/CMB2.php#L1071-L1096.

tw2113:
Opting to close this PR. If someone wants to bring the topic up again, please make sure you address all of Justin's posted backwards compatibility concerns as well with your PR.

PR #727, opened 2016-09-05
Portuguese translation pt_PT according Language Team Guidelines

pedro-mendonca:
Fixes https://github.com/WebDevStudios/CMB2/pull/709.

-  Portuguese translation according pt_PT translation team, including some of the changes proposed by @alvarogois (pt_PT GTE) in https://github.com/WebDevStudios/CMB2/pull/709, and some other updates. 

-  Added @alvarogois to contribution credits

-  Added a few newly added strings

Question: Will this translation be overriden by future Transifex imports?
Note: The translation being imported from Transifex is incomplete and isn't according pt_PT Translation Team Translation Guidelines. We are maintaining this branch and translation as the only way to keep a correct translation, discussed and according the WP core and themes/plugins translation guidelines for our locale.


jtsternberg:
Thank you for this submission, and thank you @alvarogois for #709. Now that I understand a bit more about the nuances of translations from wp.org, it will be my goal not to override translations provided through that process. Are the translations provided here, the same as [those on wp.org](https://translate.wordpress.org/locale/pt/default/wp-plugins/cmb2)? Because I cannot guarantee that the translations here will not be stomped from the translations on wp.org.

pedro-mendonca:
Hi @jtsternberg 

The stable version of the strings of CMB2 isn't still available at the wp.org repository.

The translation we've pushed here will be exactly the same that will be at wp.org as soon as it's available. When I find an error or add new strings, I usually update it on both wp.org and github (if they exist on both). 
For now, it's synced only with `Development` branch.

As this CMB2 case it's a bit different from other cases, as @alvarogois mentioned, the situation of having it bundled as a module, it's probably a good practice to keep the .po files also in github.
If you'll make the `Stable` translation set available on wp.org, you'll probably have to think on a way to sync it all, but you'll need translation editors on every language that CMB2 is translated.

PR #752, opened 2016-09-29
Taxonomy select heirarchy

### LINKED ISSUE #751: Taxonomy_select support for hierarchical Taxonomy
Related to #640 but for taxonomy_select field type. I wanted a way to show the children of a parent taxonomy term in an <optgroup> element.

So, I modified CMB2_Type_Taxonomy_Select to accept a 'hierarchy' (boolean) option. When hierarchy is true it groups the terms by parent and produces options in optgroups, e.g.:

```
<optgroup label="Parent Term">
    <option value="{term_id}" {checked}>Term name</option>
    <option value="{term_id}" {checked}>Term name</option>
</optgroup>
```

Note that this only works for taxonomies with 2 levels. `<optgroups>` can't be nested so it's not clear how what you'd do it for deeper level taxonomies anyway. If you use it with taxonomies with more levels, you'll probably get optgroups without labels (I haven't tested it).

As part of the rendering, I added CMB2_Type_Multi_Base::select_optgroup which generates the html for an optgroup. So it would be a fairly simple matter to allow optgroups on other select fields as well.

I'll submit a pull request.

I'll be happy to add documentation, but I didn't see where to do it except in the Wiki and that doesn't seem to be in any branch so I didn't want to add it until the PR is accepted.

Also of note, there is an undocumented option for the taxonomy_select field: show_option_none. It appears to allow you to change the label for the None option.

bignall:
Fixes #751 .
-  Adds hierarchy (boolean) option to taxonomy_select field
-  When hierarchy is true, it groups the taxonomy terms by parent and produces the following for each parent term.

```
<optgroup label="Parent Term">
    <option value="{term_id}" {checked}>Term name</option>
    <option value="{term_id}" {checked}>Term name</option>
</optgroup>
```

### DIFF
--- a/includes/types/CMB2_Type_Multi_Base.php
+++ b/includes/types/CMB2_Type_Multi_Base.php
@@ -13,6 +13,17 @@
 abstract class CMB2_Type_Multi_Base extends CMB2_Type_Base {
 
 	/**
+	 * Generates html for an optgroup element
+	 * @param  array $args Arguments array containing: label (label for the option group), and options (html for the options elements)
+	 * @return string Generated optgroup element html
+	 */
+	public function select_option( $args = array() ) {
+		return sprintf( "\t" . '<optgroup label="%s">' . "\n", $label) . 
+			$options .
+			sprintf( "\n\t</optgroup>\n" );
+	}
+	
+	/**
 	 * Generates html for an option element
 	 * @since  1.1.0
 	 * @param  array  $args Arguments array containing value, label, and checked boolean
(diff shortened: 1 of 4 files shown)


tw2113:
Closing as I believe a very similar option was added in 2.2.5 https://github.com/CMB2/CMB2/releases/tag/v2.2.5

`New field types, 'taxonomy_radio_hierarchical', and 'taxonomy_multicheck_hierarchical', for displaying taxonomy options in a hierarchial layout. Props to eriktelepovsky for the working code.`

If those field types aren't quite what was proposed here and the issue remains, let us know.

PR #755, opened 2016-10-01
Stop using $pagenow global

ramiy:
Replace `global $pagenow` with `$screen = get_current_screen()`

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -760,7 +760,7 @@ class CMB2 extends CMB2_Base {
 	 * @return integer $object_id Object ID
 	 */
 	public function object_id( $object_id = 0 ) {
-		global $pagenow;
+		$screen = get_current_screen();
 
 		if ( $object_id ) {
 			$this->object_id = $object_id;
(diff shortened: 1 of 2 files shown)


jtsternberg:
We cannot use get_current_screen as the context where these methods are used is not always going to be in the admin, or when that function is available.

ramiy:
Ok, lets use `get_current_screen()` on the back-end and `$pagenow` global on the front-end.

jtsternberg:
Can you state your justification for making the change?

ramiy:
Using globals is a bad practice.

WordPress core is replacing its globals (see https://core.trac.wordpress.org/ticket/37699). See changeset [38459](https://core.trac.wordpress.org/changeset/38459) and [38383](https://core.trac.wordpress.org/changeset/38383) for examples.

ramiy:
@bradp PR refreshed. Conflicts solved.

jtsternberg:
Thank you for updating. Unfortunately, https://github.com/CMB2/CMB2/pull/755#issuecomment-250936318 is still applicable and unresolved. Also, when cleaning up the merge conflicts, you removed some important elements (like the calls to `wp_unslash()`).

ramiy:
@jtsternberg Fixed.

tw2113:
@ramiy Do these account at all for people using CMB2 for frontend forms? That's part of Justin's original objections. The screen checking is very much an admin-area function, but CMB2 doesn't restrict to just there.

PR #763, opened 2016-10-06
Remove object_type default

### LINKED ISSUE #700: Term meta is saved but it doesn't appear in term edit screen later
I've run into a problem where the metadata is stored but It doesn't appear in term edit screen later.

``` php
    $oi_catalog = new_cmb2_box( array(
            'id' => $prefix . 'catalog_metabox',
            'title' => esc_attr__('Catalog settings'),
            'object_types' => array('term'),
            'taxonomies' => array('product_category'),
            'context' => 'normal',
            'priority' => 'high',
            'show_names' => true
        )
    );

    $oi_catalog_group_id = $oi_catalog->add_field( array(
            'id'          => '_cmb_catalog_repeat_group',
            'type'        => 'group',
            'options'     => array(
                'group_title'   => __( 'Banner {#}'),
                'add_button'    => __( 'Add banner'),
                'remove_button' => __( 'Remove banner'),
                'closed'    => false,
                'sortable' => false
            )
        )
    );

    $oi_catalog->add_group_field($oi_catalog_group_id, array(
            'name' => 'URL',
            'id'   => 'url',
            'type' => 'text_url'
        )
    );
```

Am I missing something? I'm running WP 4.5.3, CMB2 2.2.2.1

sbussetti:
Fixes #700.

Discussion: https://wordpress.org/support/topic/term-meta-is-saved-but-it-doesnt-appear-in-term-edit-screen-later/page/2/#post-8259773

So order matters here due to the caching pattern used in CMB2_

```
        public function object_type( $object_type = '' ) {
                if ( $object_type ) {
                        $this->object_type = $object_type;
                        return $this->object_type;
                }

                if ( $this->object_type ) {
                        return $this->object_type;
                }

                $this->object_type = $this->current_object_type();

                return $this->object_type;
        }
```

Which I think explains why different folks report different results.  Depending on the order things get initialized in, object_type could be set differently.

However, in the case where nothing else is setting object_type (this only happens during `render_form_open` and `render_form_close`), then in the above logic, current_object_type is _NEVER_ called, because $this->object_type is already defaulted to 'post'.

If you remove that default, everything works great!
-  Sets default `object_type` to null for CMB2_Base

### DIFF
--- a/includes/CMB2_Base.php
+++ b/includes/CMB2_Base.php
@@ -43,7 +43,7 @@ abstract class CMB2_Base {
 	 * @var   string
 	 * @since 2.2.3
 	 */
-	protected $object_type = 'post';
+	protected $object_type = '';
 
 	/**
 	 * Array of key => value data for saving. Likely $_POST data.


jtsternberg:
Many have reported as fixing issues with term meta not saving, so merging. Thank you for submitting.

PR #779, opened 2016-10-25
Fixed undefined variable: page_type

anhskohbo:
You're missing $page_type variable :)

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -824,7 +824,7 @@ class CMB2 extends CMB2_Base {
 			if ( is_string( $last ) ) {
 				$type = $last;
 			}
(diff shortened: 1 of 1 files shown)


tw2113:
Not quite sure what this is supposed to fix, to be honest. The original version in the line changed would return a value right away. All you're ending up doing is assigning that to a variable, inside an if statement. Never been a fan of variable assignment inside if statements myself.

Is this fixing an open issue at all? or just casual looking over of the code?

anhskohbo:
You can look `$page_type` never declared before.

tw2113:
Gotcha.

jtsternberg:
Thank you sir.

anhskohbo:
You're welcome :)

PR #789, opened 2016-11-04
Use quotes for label[for=""] selecttor

anhskohbo:
Also fixed `Syntax error, unrecognized expression:` when call CMB2 in widget (ID key has contains `[`, `]` character)


jtsternberg:
thank you!

PR #821, opened 2016-12-14
Add support for default <input type="time">

RubenMartins:
Add support for <input type="time"> as default of browsers


tw2113:
My one concern with this one is that I think it could easily be solved by using the `attributes` arguments instead of creating a brand new field type to manage.

https://github.com/CMB2/CMB2/wiki/Field-Parameters#attributes

Due to that, I'm going to close this one unless someone can provide a strong enough argument for inclusion.

We appreciate the contributions and feedback.

PR #824, opened 2016-12-20
Fork out AJAX function for re-use

wpsmith:
Need to fork out AJAX JS function so that it can be reused by extension plugins without doing more work.

### DIFF
--- a/js/cmb2.js
+++ b/js/cmb2.js
@@ -914,7 +914,7 @@ window.CMB2 = window.CMB2 || {};
 			error: function(response) {
 				cmb.log('onError', response);
 				cbs.error.call(self, response);
-			}
+			},
 			success: function(response) {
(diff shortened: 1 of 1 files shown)


jtsternberg:
hmm, the way this PR is structured with the massive white-space changes, it will be extremely hard to review. Can you make the scope of this PR very narrow, and open additional ones for the white-space changes, etc? JS hint globals definitions are not needed as they are handled in the [gruntfile](https://github.com/WebDevStudios/CMB2/blob/trunk/Gruntfile.js#L198).

tw2113:
@wpsmith any chance that you could provide a refresh of this one, seeing that the `js/cmb2.js` is finding conflicts.

@jtsternberg for what it's worth, GitHub offers the ability to ignore whitespace changes in the "diff settings" button, right next to the "review changes" button.

PR #828, opened 2017-01-07
$.closest() instead of parents()

anhskohbo:
In case I have a `li` tag wrapper `.cmb-row`, click (x) button will remove li parent.
Fixed that with closest() method.


jtsternberg:
Thanks for submitting. Can you explain to me the scenario that makes this necessary? Not sure I understand the li tag wrapper thing.

anhskohbo:
My english is not good, so pls see this gif below :)

peek 2017-01-07 11-40

jtsternberg:
This update should fix the issue. I've introduced a shared class that it targets instead of the `li` element.

anhskohbo:
Thanks!

PR #834, opened 2017-01-12
Added Select2 Post Field

mattkrupnik:

In 3rd party resources section added new field


jtsternberg:
Thank you for submitting. We already have a select2 field type highlighted on the readme:
!()[http://b.ustin.co/MZrt]

I wouldn't mind adding yours if you can outline any differences from the other one.

mattkrupnik:
Ok, I thought that this one you selected does not allow searching posts - I missed that information that **it is possible** - https://gist.github.com/mustardBees/f970b89dc45876524c8e
My bad ;(

jtsternberg:
It does, but unfortunately it's not an ajax solution. Does yours work with ajax-search?

tw2113:
Closing for the time being. If the proposed field type becomes different enough from the one we already have listed to justify, we'll re-evaluate at that time.

Thanks.

PR #838, opened 2017-01-16
Context location setup

norcross:

-  adds title display for context metaboxes with expand / collapse toggle


includes/CMB2_hookup.php
```diff
@@ -356,22 +356,20 @@ public function add_context_metaboxes() {
...
-		$title = $cmb->prop( 'title' );
+		$title = $this->cmb->prop( 'title' );
 
 		$screen = get_current_screen();
 		$hidden = get_hidden_meta_boxes( $screen );
 
-		$classes = 'postbox context-' . $cmb->prop( 'context' ) . '-box context-box ';
-		$classes .= in_array( $cmb->cmb_id, $hidden ) ? ' hide-if-js' : '';
-		$classes .= postbox_classes( $cmb->cmb_id, $screen->id );
+		$classes = 'postbox context-' . $this->cmb->prop( 'context' ) . '-box context-box postbox-container ';
```
norcross: why did you add the postbox-container class as well?

includes/CMB2_hookup.php
```diff
@@ -339,15 +339,15 @@ public function add_context_metaboxes() {
 
 				// Call our before hook.
 				if ( ! empty( $this->cmb->prop( 'title' ) ) ) {
-					$this::context_box_title_markup_open( $this->cmb );
+					$this->context_box_title_markup_open( $this->cmb );
```
norcross: if you're not going to use the `$cmb` variable passed in the function, can be removed from here

includes/CMB2_hookup.php
jtsternberg: The `postbox-container` class was being used to style the field rows, but I refactored it a bit so that it's not needed. You can see those changes in [Update context metabox output to handle no-title or no-wrap better](https://github.com/WebDevStudios/CMB2/pull/838/commits/c86b8bc0ba7c2cd020015c809fce5c76f5a1288d)

includes/CMB2_hookup.php
jtsternberg: 👍

PR #842, opened 2017-01-20
Fixed object-type detection of the `CMB2::current_object_type()` method

### LINKED ISSUE #841: The `current_object_type()` method does not resolve correctly for a metabox that has multiple values for the 'object_types' attribute when called from the 'admin-ajax.php' page context
### Expected Behavior:

For taxonomy-term based metaboxes that have multiple object-types, the field data should be persisted when called upon via `admin-ajax.php` from within the **Add New Tag** form of the `edit-tags.php` page.

### Actual Behavior:

The field data does not get saved due to the fact that the `mb_object_type()` method is used to determine the metabox's object-type based on which program execution flow is determined; the aforementioned method relies upon `current_object_type()` method that returns an invalid object-type (the default **post**) instead of the expected **term** object-type and therefore the program execution flow goes down a different path and registers the post-related hooks instead of term-related hooks and so forth.

1.  Register a taxonomy-term type metabox using the code depicted below;
2.  Navigate to the relevant taxonomy page and fill in the metabox field;
3.  Click the **Add New Tag** button and check the resulting term's field value.

### CMB2 Field Registration Code:

```php
add_action( 'cmb2_admin_init', 'yourprefix_register_demo_metabox' );
function yourprefix_register_demo_metabox() {

    $cmb = new_cmb2_box( array(
	'id'            => 'cot_test_metabox',
	'title'         => __( 'Test Metabox', 'cmb2' ),
	'object_types'  => array( 'term', 'page' ),
	'taxonomies'	=> 'cot_test_taxonomy'
    ) );

    $cmb->add_field( array(
	'name'       => __( 'Test Text', 'cmb2' ),
	'id'         => 'cot_test_field',
	'type'       => 'text',
    ) );

}

register_taxonomy('cot_test_taxonomy', 'page');
```

ADC07:
to handle the 'add-tag' action of the 'admin-ajax.php' page.

Fixes #841.

-  Add a new check to see if the current page is `admin-ajax.php` and if an **action** parameter exists within the $_POST variable determine object-type based on it's value.

PS: AFAICT, there is only one entry point into the plugin's execution flow via the 'admin-ajax.php' page and this applies to this specific scenario / **term** object-type therefore the **action** parameter validation is sort of useless however, perhaps this might change in the near future so I'm recommending we leave it there.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -953,6 +953,14 @@ class CMB2 extends CMB2_Base {
 			$type = 'term';
 		}
 
+		if ( $pagenow === 'admin-ajax.php' && isset($_POST['action']) ) {
+		    switch ( $_POST['action'] ) {
+			case 'add-tag':
+			    $type = 'term';
+			    break;
+		    }
+		}
+
 		return $type;
 	}
 


jtsternberg:
This is great, and the bug report, steps for reproducing, etc were all very clear, and the PR is focused. This makes for an easy merge. Thank you!

PR #874, opened 2017-03-08
Update "Image" Definition

kuchenundkakao:
If SVG Upload is enabled, SVG can't be chosen in file_list Field because it is MIME-Type "image", but doesn't have Size Data.

### DIFF
--- a/js/cmb2.js
+++ b/js/cmb2.js
@@ -265,7 +265,7 @@ window.CMB2 = window.CMB2 || {};
 		};
 
 		handlers.getAttachmentHtml = function( attachment, templatesId ) {
-			var isImage = 'image' === attachment.get( 'type' );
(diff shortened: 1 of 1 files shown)


jtsternberg:
There needs to be another way to determine this. Currently, this PR causes images to be processed as files: http://b.ustin.co/lg5j

kuchenundkakao:
yeah sorry, i thought i could use the wp global methods. Replaced it with a check for the svg extension for now.

jtsternberg:
Played with this a bit. I think it would be more ideal to use the svg as an image and output the preview (and workaround the "size" checks).

jtsternberg:
Please test this branch as I've attempted to fix the svg image display: https://github.com/WebDevStudios/CMB2/tree/fix-svg-display

kuchenundkakao:
Yep, Works ;)

PR #885, opened 2017-03-17
Get PHP7 builds to pass on Travis again.

desrosj:
WordPress' test suite fails on PHP7 in Travis currently because of the version of PHPUnit (6) Travis includes on their PHP7.x builds.

[Core ticket 40086](https://core.trac.wordpress.org/ticket/40086) fixes this until [WordPress' test suite is updated to support PHPUunit 6](https://core.trac.wordpress.org/ticket/39822).

This pull should pass on PHP7 instead of allowing it to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,12 +30,15 @@ matrix:
       php: 7.0
     - env: WP_VERSION=3.8 WP_MULTISITE=1
       php: 7.0
-    - env: WP_VERSION=latest WP_MULTISITE=0
-      php: 7.0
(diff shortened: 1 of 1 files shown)


tw2113:
@jtsternberg similar changes worked well for CPTUI, so I can confirm that it's working as a whole here.

desrosj:
It might be worth adding a comment around these lines, or opening a separate issue linking to the Ticket on Trac for PHPUnit 6 support so that this is revisited when that gets closed. This can be removed then.

jtsternberg:
Nice work, and thank you!

PR #905, opened 2017-04-03
Fix WP VIP PHPCS in css/index

bradp:
Newline


jtsternberg:
Hoping wp phpcs exceptions this... we'll see if any movement.

bradp:
There were only a handful of these, so not a huge issue of they don't.

jtsternberg:
might as well.

PR #946, opened 2017-04-04
Before/after field row hooks

### LINKED ISSUE #876: A way to add content before/after field based on field arg
Revising documentation about how to create a custom fields and CMB2 code I am trying to setup a plugin that renders html before and after of a field if this has a parameter(s) defined

I am trying to do the same that wordpress publishing actions form (before publish button) with any CMB2 field, the workflow is really easy but I need to find a way to add content before/after any field

The problem is there is any hook before/after a field, so at this moment is impossible to make work correctly this plugin

Can someone with more deep knowledge with CMB2 help me?

rubengc:
Fixes #876

Added
cmb2_before_field_row
cmb2_before_{$field_type}_field_row
cmb2_after_field_row
cmb2_after_{$field_type}_field_row

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -324,6 +324,18 @@ class CMB2 extends CMB2_Base {
 		/**
 		 * Hook after form form has been rendered
 		 *
+		 * @param array  $cmb_id      The current box ID.
+		 * @param int    $object_id   The ID of the current object.
+		 * @param string $object_type The type of object you are working with.
+		 *	                           Usually `post` (this applies to all post-types).
+		 *	                           Could also be `comment`, `user` or `options-page`.
+		 * @param array  $cmb         This CMB2 object.
+		 */
+		do_action( 'cmb2_after_form', $this->cmb_id, $object_id, $object_type, $this );
+
+		/**
+		 * Hook after form form has been rendered
+		 *
 		 * The dynamic portion of the hook name, $this->cmb_id, is the meta_box id.
 		 *
 		 * The first dynamic portion of the hook name, $object_type, is the type of object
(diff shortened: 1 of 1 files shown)


rubengc:
IMPORTANT!

Writing this PR I noticed a wrong hooks order that needs to be solved at CMB2.php

Hooks are called in this order:
`cmb2_before_form`
`cmb2_before_{$object_type}_form_{$cmb_id}`
`cmb2_after_form`
`cmb2_after_{$object_type}_form_{$cmb_id}`

And they should be called at this order:
`cmb2_before_form`
`cmb2_before_{$object_type}_form_{$cmb_id}`
`cmb2_after_{$object_type}_form_{$cmb_id}`
`cmb2_after_form`

Why? Because this means call this hooks is really dangerous because you could break your html if you expect a relative order, for example, if two plugins using this hooks to print something, the most common workflow is use same relative callbacks (`cmb2_before_form` and `cmb2_after_form`) or (`cmb2_before_{$object_type}_form_{$cmb_id}` and `cmb2_after_{$object_type}_form_{$cmb_id}`) 

A visual example with two plugins (Plugin 1 and Plugin 2):
Current order:
`cmb2_before_form` Plugin 1 opens a div
`cmb2_before_{$object_type}_form_{$cmb_id}` Plugin 2 prints a p
`cmb2_after_form`  Plugin 1 closes the div
`cmb2_after_{$object_type}_form_{$cmb_id}` Plugin 2 closes the p

Result:
``` html
<div>
<p>
</div>
</p>
```

New order:
`cmb2_before_form` Plugin 1 opens a div
`cmb2_before_{$object_type}_form_{$cmb_id}` Plugin 2 prints a p
`cmb2_after_{$object_type}_form_{$cmb_id}` Plugin 2 closes the p
`cmb2_after_form`  Plugin 1 closes the div

Result:
``` html
<div>
<p>
</p>
</div>
```

I keep the same order as before/after form to this PR because I do not know if this is the desired order, anyway if you confirm that you want to change this order, I will update this PR too

jtsternberg:
> ... wrong hooks order that needs to be solved at CMB2.php

This is an excellent point. Would appreciate a separate PR to resolve that order, if you are up for it. If so, please also a changelog entry, similar to this: https://github.com/CMB2/CMB2/commit/a03b471441172a7ba74811f45792872682bf967b (since it is a backwards-compatibility concern).

rubengc:
New hooks order to match coming CMB2 PR:
`cmb2_before_field_row`
`cmb2_before_{$field_type}field_row`
`cmb2_after{$field_type}_field_row`
`cmb2_after_field_row`

Ok @jtsternberg I will PR it in a separate PR with changelog notes

jtsternberg:
RE: this PR, i'm not super sure this is the way to handle this. Ideally, we could do something with `CMB2_Field:peform_param_callback()`.

rubengc:
Dat github desktop :/

What do you mean @jtsternberg ?

rubengc:
I am not able to open a new PR with this field row hooks and the amount of reverts generated from github desktop app, so on get this pulled (or declined) I will folk again CMB2 repo

rubengc:
No news about this?

PR #949, opened 2017-04-04
WPCS VIP sniffs -- includes/CMB2.php

jazzsequence:
Fixes (most) WPCS VIP sniffs. Sniffs that were not addressed:

```
    1 | ERROR   | [ ] Filenames should be all lowercase with hyphens
      |         |     as word separators. Expected cmb2.php, but
      |         |     found CMB2.php.
    1 | ERROR   | [ ] Class file names should be based on the class
      |         |     name with "class-" prepended. Expected
      |         |     class-cmb2.php, but found CMB2.php.
```

...and variations of:

```
  906 | WARNING | [ ] Detected access of super global var $_REQUEST,
      |         |     probably needs manual inspection.
 1028 | WARNING | [ ] Detected access of super global var $_POST,
      |         |     probably needs manual inspection.
```

Tested locally to pass unit tests but should have a sanity check.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -492,7 +492,7 @@ class CMB2 extends CMB2_Base {
 		$closed_class = $field_group->options( 'closed' ) ? ' closed' : '';
 
 		echo '
-		<div class="postbox cmb-row cmb-repeatable-grouping', $closed_class, '" data-iterator="', $field_group->index, '">';
+		<div class="postbox cmb-row cmb-repeatable-grouping', wp_kses( $closed_class, array() ), '" data-iterator="', absint( $field_group->index ), '">';
 
 		if ( $field_group->args( 'repeatable' ) ) {
 			echo '<button type="button" ', $remove_disabled, 'data-selector="', $field_group->id(), '_repeat" class="dashicons-before dashicons-no-alt cmb-remove-group-row" title="', esc_attr( $field_group->options( 'remove_button' ) ), '"></button>';
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -888,7 +888,7 @@ class CMB2 extends CMB2_Base {
 		// Try to get our object ID from the global space.
 		switch ( $this->object_type() ) {
 			case 'user':
-				$object_id = isset( $_REQUEST['user_id'] ) ? wp_unslash( absint( $_REQUEST['user_id'] ) ) : $object_id;
+				$object_id = ( wp_verify_nonce( $this->nonce() ) && isset( $_REQUEST['user_id'] ) ) ? wp_unslash( absint( $_REQUEST['user_id'] ) ) : $object_id;
 				$object_id = ! $object_id && 'user-new.php' !== $pagenow && isset( $GLOBALS['user_ID'] ) ? $GLOBALS['user_ID'] : $object_id;
 				break;
 
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -1450,6 +1450,13 @@ class CMB2 extends CMB2_Base {
 		return $this->prop( 'context' ) && in_array( $this->prop( 'context' ), array( 'form_top', 'before_permalink', 'after_title', 'after_editor' ), true );
 	}
 
+	public function kses_group_wrap_atts() {
+		return array(
+			'class' => array(),
+			'style' => array(),
+		);
+	}
+
 	/**
 	 * Magic getter for our object.
 	 *
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -889,7 +889,7 @@ class CMB2 extends CMB2_Base {
 		// Try to get our object ID from the global space.
 		switch ( $this->object_type() ) {
 			case 'user':
-				$object_id = isset( $_REQUEST['user_id'] ) ? wp_unslash( $_REQUEST['user_id'] ) : $object_id;
+				$object_id = isset( $_REQUEST['user_id'] ) ? wp_unslash( absint( $_REQUEST['user_id'] ) ) : $object_id;
 				$object_id = ! $object_id && 'user-new.php' !== $pagenow && isset( $GLOBALS['user_ID'] ) ? $GLOBALS['user_ID'] : $object_id;
 				break;
 
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -436,7 +436,7 @@ class CMB2 extends CMB2_Base {
 		}
 
 		if ( $field_group->args( 'repeatable' ) ) {
-			echo '<div class="cmb-row"><div class="cmb-td"><p class="cmb-add-row"><button type="button" data-selector="', esc_attr( $field_group->id() ), '_repeat" data-grouptitle="', esc_attr( $field_group->options( 'group_title' ) ), '" class="cmb-add-group-row button-secondary">', $field_group->options( 'add_button' ), '</button></p></div></div>';
+			echo '<div class="cmb-row"><div class="cmb-td"><p class="cmb-add-row"><button type="button" data-selector="', esc_attr( $field_group->id() ), '_repeat" data-grouptitle="', esc_attr( $field_group->options( 'group_title' ) ), '" class="cmb-add-group-row button-secondary">', esc_html( $field_group->options( 'add_button' ) ), '</button></p></div></div>';
 		}
 
 		echo '</div></div></div>';
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -87,14 +87,14 @@ class CMB2 extends CMB2_Base {
  		 * and handles hooking in and saving those fields.
 		 */
 		'hookup'           => true,
-		'save_fields'      => true, // Will not save during hookup if false.
+		'save_fields'      => true,  // Will not save during hookup if false.
 		'closed'           => false, // Default metabox to being closed.
 		'taxonomies'       => array(),
 		'new_user_section' => 'add-new-user', // or 'add-existing-user'.
 		'new_term_section' => true,
 		'show_in_rest'     => false,
-		'classes'          => null, // Optionally add classes to the CMB2 wrapper
-		'classes_cb'       => '', // Optionally add classes to the CMB2 wrapper (via a callback)
+		'classes'          => null,  // Optionally add classes to the CMB2 wrapper
+		'classes_cb'       => '',    // Optionally add classes to the CMB2 wrapper (via a callback).
 	);
 
 	/**
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -1450,11 +1450,24 @@ class CMB2 extends CMB2_Base {
 		return $this->prop( 'context' ) && in_array( $this->prop( 'context' ), array( 'form_top', 'before_permalink', 'after_title', 'after_editor' ), true );
 	}
 
-	public function kses_group_wrap_atts() {
-		return array(
-			'class' => array(),
-			'style' => array(),
-		);
+	public function kses_args( $type ) {
+		$params = array();
+
+		switch ( $type ) {
+			case 'group_wrap_attributes' :
+				$params = array(
+					'class' => array(),
+					'style' => array(),
+				);
+				break;
+			case 'remove_disabled' :
+				$params = array(
+					'disabled' => array(),
+				);
+		}
+
+
+		return $params;
 	}
 
 	/**
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -1297,7 +1297,9 @@ class CMB2 extends CMB2_Base {
 	 */
 	protected function _add_field_to_array( $field, &$fields, $position = 0 ) {
 		if ( $position ) {
-			CMB2_Utils::array_insert( $fields, array( $field['id'] => $field ), $position );
+			CMB2_Utils::array_insert( $fields, array(
+				$field['id'] => $field,
+			), $position );
 		} else {
 			$fields[ $field['id'] ] = $field;
 		}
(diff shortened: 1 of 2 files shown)


jtsternberg:
Reviewing too. Let's move `kses_args()` to `CMB2_Base`, so it's accessible to `CMB2_Field`, etc.

includes/CMB2.php
```diff
@@ -1449,6 +1455,43 @@ public function is_alternate_context_box() {
...
+	}
+
+	/**
+	 * Helper function that returns whether a field group is repeatable or not.
+	 *
+	 * @since  2.2.4
+	 * @param  object $field_group The field group.
+	 * @return boolean
+	 */
+	public function is_repeatable( $field_group ) {
```
jtsternberg: This seems like an odd method to add. What is the purpose?
`$this->is_repeatable( $field_group );`
`$field_group->args( 'repeatable' );`

jazzsequence:
So the nonce checks were pulled out in https://github.com/CMB2/CMB2/pull/949/commits/4e7723ab060f41899a69c4bb410ca15f607c4976 and those will fail sniff tests, too, but it's not as if we aren't checking nonces or making sure the data we're getting from superglobals isn't sane, so ¯\_(ツ)_/¯.

includes/CMB2_Base.php
```diff
@@ -379,6 +379,43 @@ public static function maybe_hook( $val, $hook_name, $hook_function ) {
...
+	}
+
+	/**
+	 * Helper function that returns whether a field group is repeatable or not.
+	 *
+	 * @since  2.2.4
+	 * @param  object $field_group The field group.
+	 * @return boolean
+	 */
+	public function is_repeatable( $field_group ) {
```
jtsternberg: This seems like an odd micro-optimization. Is there a reason for this method?
`$this->is_repeatable( $field_group );`
vs
`$field_group->args( 'repeatable' );`

jtsternberg: the only way this feels better to me is if this is moved to `CMB2_Field`, and even then, I could see creating an `CMB2_Field::is( $param )` method that allows checking boolean field parameters (so `$field_group->args( 'repeatable' );` would become `$field_group->is( 'repeatable' );`).

includes/CMB2.php
```diff
@@ -410,16 +412,16 @@ public function render_group( $args ) {
 
 		$field_group->peform_param_callback( 'before_group' );
 
-		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', $this->group_wrap_attributes( $field_group ), '>';
+		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', wp_kses( $this->group_wrap_attributes( $field_group ), $this->kses_args( 'group_wrap_attributes' ) ), '>';
```
jazzsequence: Is it possible to unit test that sort of back-compat stuff? I'd rather have extensive -- even overkill -- unit tests that can tell us verifiably what is likely to break than have to stare at things and hypothesize or run tests manually against really old versions of CMB2 and I'm guessing that would state our case better than "we can't not sanitize this because it could break backwards compatibility" when presenting it to VIP.

includes/CMB2.php
```diff
@@ -410,16 +412,16 @@ public function render_group( $args ) {
...
 			$class = $desc ? ' cmb-group-description' : '';
-			echo '<div class="cmb-row', $class, '"><div class="cmb-th">';
+			echo '<div class="cmb-row', wp_kses( $class, array() ), '"><div class="cmb-th">';
 			if ( $label ) {
-				echo '<h2 class="cmb-group-name">', $label, '</h2>';
+				echo '<h2 class="cmb-group-name">', wp_kses_post( $label ), '</h2>';
 			}
 			if ( $desc ) {
-				echo '<p class="cmb2-metabox-description">', $desc, '</p>';
+				echo '<p class="cmb2-metabox-description">', wp_kses_post( $desc ), '</p>';
```
jazzsequence: For things like this, though, where there's an obvious usage (it's a description, so describe things) and a fairly obvious _mis_ use (injecting arbitrary scripts), doesn't it make sense to draw a line in the sand and say "yeah, that's not a good idea"? If anyone wonders why the 200 lines of javascript they added to their description is getting stripped out, it should be fairly obvious. And a solution would be to build their own field type where they can put as much arbitrary javascript in there as they want.

includes/CMB2.php
jtsternberg: Yes, extensive unit tests should def. be written for this, though i'm not sure I 100% follow what you're saying as far as what should be tested. Let me write some tests for how it can/should work now, and we can go from there.

includes/CMB2_Base.php
jazzsequence: `$this->is_repeatable( $field_group )` is there for a couple reasons, and I'm happy to pull it out.

1. Scrutinizer was complaining about the repetition of `$field_group->args( 'repeatable' )` and it was done in an attempt to "leave the code better than you found it."
2. To Scrutinizer's point, it's done a whole bunch of times, and anything that's done more than once can be split into it's own function (admittedly, the function is just a single line that spits out the same output, so the value is questionable except...)
3. It's a couple less characters to type.

Ideally, the way I originally conceived it, it would be more global like `is_repeatable()` instead of being a class method that requires a parameter, but that gets hairy, too.

I like the `CMB2_Field::is( $param )` and being able to use it as `$field_group->is( 'repeatable' )` idea, though. It seems like an obvious evolution of the same idea.

jazzsequence:
>In general, I'd much prefer if all the stylistic, non-functional changes were in a separate PR. Will make it much easier to review both PRs, and to communicate about functional changes w/o cruft of the other stuff.

Good point. I was working through the results of the sniff tests and wasn't 100% sure where it would take me. It makes sense to break stylistic stuff out into its own thing though. I will leave this open for now (just because there's discussion and to refer back to later), but I can ultimately create separate PRs for each and then close this PR.

includes/CMB2.php
jazzsequence: Basically, I'd rather know I can rely on the unit tests to tell me if something is going to break backwards compatibility than stare at it and guess (or test it manually myself). 

I know that it's a lot of assumptions, like _maybe_ someone decides to throw javascript in somewhere where it doesn't belong or whatever, but if that's a legitimate concern, let's test against it, so the tests can fail, we can figure out a sanitization solution that _works_ and then move on. And for some things I do think that a line in the sand needs to be drawn where the data makes sense.

jtsternberg: line in the sand (and tests) created: https://github.com/CMB2/CMB2/commit/fec5acbe8890afd3f4567897a9bf596c6cbe573e

includes/CMB2.php
jtsternberg: I think that makes sense. good call.

includes/CMB2_Base.php
jtsternberg: > It's a couple less characters to type.

but... it's not. :)
image

jazzsequence: uggggghhhh
Well it _looks_ like fewer characters, okay?
I stand corrected :)

includes/CMB2.php
```diff
@@ -490,15 +492,15 @@ public function render_group_row( $field_group, $remove_disabled ) {
...
-		if ( $field_group->args( 'repeatable' ) ) {
-			echo '<button type="button" ', $remove_disabled, 'data-selector="', $field_group->id(), '_repeat" class="dashicons-before dashicons-no-alt cmb-remove-group-row" title="', esc_attr( $field_group->options( 'remove_button' ) ), '"></button>';
+		if ( $this->is_repeatable( $field_group ) ) {
+			echo '<button type="button" ', wp_kses( $remove_disabled, $this->kses_args( 'remove_disabled' ) ), 'data-selector="', esc_attr( $field_group->id() ), '_repeat" class="dashicons-before dashicons-no-alt cmb-remove-group-row" title="', esc_attr( $field_group->options( 'remove_button' ) ), '"></button>';
 		}
 
 			echo '
 			<div class="cmbhandle" title="' , esc_attr__( 'Click to toggle', 'cmb2' ), '"><br></div>
-			<h3 class="cmb-group-title cmbhandle-title"><span>', $field_group->replace_hash( $field_group->options( 'group_title' ) ), '</span></h3>
+			<h3 class="cmb-group-title cmbhandle-title"><span>', wp_kses_post( $field_group->replace_hash( $field_group->options( 'group_title' ) ) ), '</span></h3>
```
mikeselander: Do you think we should allow any HTML in here? One could easily break proper HTML structure since this is within a `span` but not sure how people actually use this.

includes/CMB2.php
```diff
@@ -410,16 +412,16 @@ public function render_group( $args ) {
...
-		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', $this->group_wrap_attributes( $field_group ), '>';
+		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', wp_kses( $this->group_wrap_attributes( $field_group ), $this->kses_args( 'group_wrap_attributes' ) ), '>';
 
 		if ( $desc || $label ) {
 			$class = $desc ? ' cmb-group-description' : '';
-			echo '<div class="cmb-row', $class, '"><div class="cmb-th">';
+			echo '<div class="cmb-row', wp_kses( $class, array() ), '"><div class="cmb-th">';
 			if ( $label ) {
-				echo '<h2 class="cmb-group-name">', $label, '</h2>';
+				echo '<h2 class="cmb-group-name">', wp_kses_post( $label ), '</h2>';
```
mikeselander: @jtsternberg should someone ever be putting HTMl into this title? Looks like this is just the name attribute so no need for HTML.

includes/CMB2.php
```diff
@@ -410,16 +412,16 @@ public function render_group( $args ) {
 
 		$field_group->peform_param_callback( 'before_group' );
 
-		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', $this->group_wrap_attributes( $field_group ), '>';
+		echo '<div class="cmb-row cmb-repeat-group-wrap ', esc_attr( $field_group->row_classes() ), '" data-fieldtype="group"><div class="cmb-td"><div data-groupid="', esc_attr( $field_group->id() ), '" id="', esc_attr( $field_group->id() ), '_repeat" ', wp_kses( $this->group_wrap_attributes( $field_group ), $this->kses_args( 'group_wrap_attributes' ) ), '>';
 
 		if ( $desc || $label ) {
 			$class = $desc ? ' cmb-group-description' : '';
-			echo '<div class="cmb-row', $class, '"><div class="cmb-th">';
+			echo '<div class="cmb-row', wp_kses( $class, array() ), '"><div class="cmb-th">';
```
jazzsequence: It's mostly to shut the sniff up and I don't know how VIP would feel about adding whitelisting comments to things if they found them... (guessing not very good).

includes/CMB2.php
jazzsequence: I can see a situation where someone might do something like 

     <h2 class="cmb-group-name">Something <em>really</em> important</h2>

In which case you'd want to make sure the HTML was okay.

includes/CMB2.php
jazzsequence: Ditto above.

includes/CMB2.php
mikeselander: Good point 👍 

Should we at least pair down the list of available options to ones that are valid within a `<h2>` tag?

includes/CMB2_Base.php
```diff
@@ -379,6 +379,43 @@ public static function maybe_hook( $val, $hook_name, $hook_function ) {
...
+		switch ( $type ) {
+			case 'group_wrap_attributes':
+				$params = array(
+					'class' => array(),
+					'style' => array(),
+				);
+				break;
+			case 'remove_disabled':
+				$params = array(
+					'disabled' => array(),
```
jazzsequence: Partially because I see this function expanding to various wp_kses arguments for different things we need to sanitize throughout the code, so ideally there would be one place for all the things.

includes/CMB2.php
mikeselander: Well, that would trigger a manual review of this line. Explanatory comments are fine but we should probably keep this list short for anyone looking. No real problem with running it through `esc_attr` I suppose.

includes/CMB2.php
jtsternberg: The name field param is what is generally stuffed in the `<label>`, but in the group field's case, it's output above the group (and yikes, looks like some css needs to be fixed for it), but yah, as @jazzsequence mentioned, the 'name' params should probably allow some html

jazzsequence: Re: paring down the list, we _could_ do that -- heading tags, for example, probably shouldn't be inside a heading tag -- but I think `wp_kses_post` is a quick and dirty solution that we can implement sooner and then refine it via `kses_args` later.

tw2113:
Going to close. We can still read/review the original PR and break out as necessary then.

PR #953, opened 2017-04-13
Before/After row hooks

### LINKED ISSUE #876: A way to add content before/after field based on field arg

rubengc:
Same PR as #946

Fixes #876 

Added
cmb2_before_field_row
cmb2_before_{$field_type}field_row
cmb2_after_{$field_type}_field_row
cmb2_after_field_row


rubengc:
ping @jtsternberg

tw2113:
@rubengc I realize it's been awhile. Our apologies there.

What is the difference between these and the options available with https://github.com/CMB2/CMB2/wiki/Field-Parameters#before-after-before_row-after_row-before_field-after_field ? All of those receive the field args and field object which very closely matches what you're doing with your PR.

PR #954, opened 2017-04-13
Before and after form hooks order

rubengc:
This is a backwards-compatibility break. If your theme or plugin is using some of next action hooks and your are connecting the output between them then take a look a this explanation

Hooks have been called in this order (until this PR):
`cmb2_before_form`
`cmb2_before_{$object_type}_form_{$cmb_id}`
`cmb2_after_form`
`cmb2_after_{$object_type}_form_{$cmb_id}`

And now they are called at this order:
`cmb2_before_form`
`cmb2_before_{$object_type}_form_{$cmb_id}`
`cmb2_after_{$object_type}_form_{$cmb_id}`
`cmb2_after_form`

Why? Because this means call this hooks is really dangerous because you could break your html if you expect a relative order, for example, if two plugins using this hooks to print something, the most common workflow is use same relative callbacks (`cmb2_before_form` and `cmb2_after_form`) or (`cmb2_before_{$object_type}_form_{$cmb_id}` and `cmb2_after_{$object_type}_form_{$cmb_id}`) 

A visual example with two plugins (Plugin 1 and Plugin 2):
Current order:
`cmb2_before_form` Plugin 1 opens a div
`cmb2_before_{$object_type}_form_{$cmb_id}` Plugin 2 prints a p
`cmb2_after_form`  Plugin 1 closes the div
`cmb2_after_{$object_type}_form_{$cmb_id}` Plugin 2 closes the p

Result:
``` html
<div>
<p>
</div>
</p>
```

New order:
`cmb2_before_form` Plugin 1 opens a div
`cmb2_before_{$object_type}_form_{$cmb_id}` Plugin 2 prints a p
`cmb2_after_{$object_type}_form_{$cmb_id}` Plugin 2 closes the p
`cmb2_after_form`  Plugin 1 closes the div

Result:
``` html
<div>
<p>
</p>
</div>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ All notable changes to this project will be documented in this file.
 * New action hook, `cmb2_footer_enqueue`, which occurs after CMB2 enqueues its assets.
 * Example functions clean up. Props [@PavelK27](https://github.com/PavelK27) ([#866](https://github.com/CMB2/CMB2/pull/866)).
 * New `CMB2_Utils` methods, `get_available_image_sizes()` and `get_named_size()`. Props [@Cai333](https://github.com/Cai333). 
+* New before/after form hooks order. For more details see ([#954](https://github.com/CMB2/CMB2/pull/954)).
 
 ### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


rubengc:
About this PR, I just did a plugin for cmb2 to add tabs that uses this hooks to output div open/close tags

https://github.com/rubengc/cmb2-tabs

jtsternberg:
Thank you!

PR #973, opened 2017-05-26
Add the ability to use keys in group title

schrapel:
This PR introduces the ability to add field values into the group_title. For example

```
'options' => [
  'group_title'   => __( 'User {user_id} with roles: {user_roles}'  ),
  'add_button'    => __( 'Add Another User', 'cmb2' ),
  'remove_button' => __( 'Remove User', 'cmb2' ),
],
```

It will work for any fields that have a value of string or an array. If there are fields with other values it will ignore these. We could add more fields in the future but I think these are the main two people would want.

The reason for this change is that we have a repeatable field that has 100 entries and each field takes up the majority of the screen. With this change we can have the fields collapsed and an overview is shown for each collapsed field on what the values inside the field contain.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -547,7 +547,7 @@ class CMB2 extends CMB2_Base {
 
 			echo '
 			<div class="cmbhandle" title="' , esc_attr__( 'Click to toggle', 'cmb2' ), '"><br></div>
-			<h3 class="cmb-group-title cmbhandle-title"><span>', $field_group->replace_hash( $field_group->options( 'group_title' ) ), '</span></h3>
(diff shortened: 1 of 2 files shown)


jtsternberg:
This is a neat concept, but since there is no JS equivalent, it is half-broken from the start. Once you make any changes to the ordering or add groups, this starts to get wonky.

jtsternberg:
If you're simply looking for something custom, you could adapt this snippet to update the group titles to whatever you want: https://github.com/CMB2/CMB2-Snippet-Library/blob/master/javascript/dynamically-change-group-field-title-from-subfield.php

schrapel:
Thanks for taking a look. JS does seem like a better solution here.

PR #979, opened 2017-06-06
Fix spacing issues.

xBLADEx:
Fixes #{issue-number}.

-  Fix the linting errors for spacing.

-  

-


tw2113:
Some thoughts:

Changes should be PR'd against trunk, not master.
If the PR still ends up being notably big even against the correct branch, it will be easier to PR in chunks instead of one huge one that is harder to review.

tw2113:
Quick followup, this may be the only real change? https://github.com/CMB2/CMB2/pull/979/commits/1ba8779de696b3ed4db19c6a2c18cbcca9d93b5e

jtsternberg:
Agree with @tw2113. Please issue a clean PR against trunk.

PR #993, opened 2017-07-05
Cast variable

anhskohbo:
The method just return cast `$this->options` as array but actually `$this->options` assign before still a string in some case.

Bugs will be show when I try empty cmb setting in `wp_options` table.

### DIFF
--- a/includes/CMB2_Options.php
+++ b/includes/CMB2_Options.php
@@ -200,14 +200,14 @@ class CMB2_Option {
 			$test_get = apply_filters( "cmb2_override_option_get_{$this->key}", 'cmb2_no_override_option_get', $default, $this );
 
 			if ( 'cmb2_no_override_option_get' !== $test_get ) {
-				$this->options = $test_get;
+				$this->options = (array) $test_get;
 			} else {
 				// If no override, get the option
-				$this->options = get_option( $this->key, $default );
+				$this->options = (array) get_option( $this->key, $default );
 			}
 		}
 
-		return (array) $this->options;
+		return $this->options;
 	}
 
 }
--- a/tests/test-cmb-core.php
+++ b/tests/test-cmb-core.php
@@ -314,6 +314,29 @@ class Test_CMB2_Core extends Test_CMB2 {
 		$this->assertEquals( $val, $new_value );
 	}
 
+	public function test_cmb2_with_empty_options() {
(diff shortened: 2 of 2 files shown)


jtsternberg:
You will need to write unit tests to prove/test your scenario.

anhskohbo:
Added unit tests

jtsternberg:
I just ran the tests you wrote with and without the changes you made in `includes/CMB2_Options.php`. To properly test your changes, at least 1 of your assertions should fail without your patch. Otherwise, there is no reason to make the change. Make sense?

jtsternberg:
Thank you for your PR. I have actually merged it to get your unit tests (and [given you props](https://github.com/CMB2/CMB2/commit/6cb60c1842e22b419ad4b219ec548c29f959316a)), but I reverted the [`Cast variable`](https://github.com/CMB2/CMB2/pull/993/commits/95c4ccaa20e975e4eb73d41baab98250e539c1fe) changes, as they are actually causing incorrect results. I have written another unit test (and moved them all to [`tests/test-cmb-options.php`](https://github.com/CMB2/CMB2/blob/trunk/tests/test-cmb-options.php)) that asserts that an empty option value is still cast to an array (as it should be). When I apply your patch, those tests break, so I'm not sure what the intention of your patch, but it breaks the way `CMB2_Options` should work.

anhskohbo:
I miss forgot line in last commit:

```php
public function setUp() {
    ...

    add_option( 'cmb_empty_option', '' );
}
```
See the unit tests failures:

peek 2017-07-08 00-02

The bug is because this `CMB2_Option::get_options`:

```php
$this->options = get_option( $this->key, $default );

...

return (array) $this->options;
```

and on `CMB2_Option::update()` or `CMB2_Option::remove()` method:

```php
public function remove( $field_id, $resave = false ) {

	$this->get_options();

	if ( isset( $this->options[ $field_id ] ) ) {
		unset( $this->options[ $field_id ] );
	}
```

If some reasons, I have a option with empty string (or nothing value). And call `CMB2_Option::remove()` will be given error `Illegal string offset ...`.

Because you only cast when return options. Actually `$this->options` still is a string.

jtsternberg:
Ah, now I understand! ed2e58f should address that specific issue. Thank you for providing more info/clarification.

PR #994, opened 2017-07-07
Updates 'cmb2_can_save' filter

Arno33:

-  Adding the current CMB2 box object as a parameter to the 'cmb2_can_save' filter so we can identify which box is being saved.

-  It allows us to easily get and verify the box nonce into our custom filter for example.


tw2113:
I see no major issue with passing the extra parameter. @jtsternberg any issue from you?

PR #996, opened 2017-07-11
add apply_filters for the output html row

polupraneeth:
It filter will help create tabs and have more control on output html row - https://github.com/stackadroit/cmb2-extensions

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -824,8 +824,8 @@ class CMB2_Field extends CMB2_Base {
 
 			$this->peform_param_callback( 'before_row' );
 			
-			// Ok, callback is good, let's run it and store the result.
-        		ob_start();
+			// Ok, let's run it and store the result.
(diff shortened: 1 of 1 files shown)


tw2113:
I'd argue it'd be best to turn on the output buffering after the potential early returns. Would be curious to see how turning that on, and then returning early, affected things, if at all.

polupraneeth:
Hi, @tw2113 sorry I am not to good in English, unable to understand your comment. Should i have to make some changes ?

tw2113:
Basically move line 814, the `ob_start();`, to line 825

polupraneeth:
Done, moved the  ob_start() to line 825 :)

jtsternberg:
You can already override the render callback: https://github.com/CMB2/CMB2/blob/develop/example-functions.php#L200
Closing unless a compelling argument can be made for needing this filter instead of using the available field parameter, `render_row_cb`.

PR #1003, opened 2017-07-28
Fix JS to prevent meta keys with | or / from breaking file fields

lipemat:
v2.2.4 as well as tested against trunk.
Tested a few WP versions with same results.

**Issue**
Using a meta_key (id) for a field that is namespaced with / or | breaks the remove  JS (cmb.handleRemoveMedia()) for the file field.

**To Recreate:**

1.  Add a field with the type of file
2. Use a / somewhere in the id (e.g. cmb2/meta/featured)
3. Add a file to this field
4. Click the remove icon
5. View the browser console to see the error. Also notice that file is not removed.


jtsternberg:
thanks!

jtsternberg:
In the future, please submit pull requests to the trunk branch.

PR #1013, opened 2017-08-15
Fix select value floats

### LINKED ISSUE #869: Leading zero values not respected
If a select is populated with, say, months, where leading zeros are added to numeric values (e.g. December is `12` but January is `01`) the correct value should be selected after saving.

However, this isn't the case. I think it's these lines in 2.2.4, `includes/types/CMB2_Type_Multi_Base.php:77`:

		if ( is_numeric( $value ) ) {
			$value = intval( $value );
		}

Is there a reason for this?

jtsternberg:
Fixes #869

This will need testing (preferably by those who reported in #869).


jtsternberg:
This has been merged.

PR #1015, opened 2017-08-16
Restore `wp_unslash()` for money values

### LINKED ISSUE #1014: Missing wp_unslash() for money values
In Swiss German the thousand separator is an apostrophe so a money value can be `2'180.00`. But if it gets sanitized the value of `$this->value` is actually `2\'180.00` which breaks the replacements and you get `2.00`.

This seems to be introduced by c325d63aee8c36095c66564d56d24c40f96ccd99.

ocean90:
Fixes #1014.


jtsternberg:
Would you be able to add a test to supplement [the existing slashes test](https://github.com/CMB2/CMB2/blob/trunk/tests/test-cmb-field.php#L522-L543)?

jtsternberg:
Thank you sir.

ocean90:
Thanks for the test. 👍

PR #1026, opened 2017-09-14
Fixes invalid call to CMB2_Utils::filter_empty

Sharsie:
When the repeatable group field is empty (all groups are removed) a null value is sent to CMB2_Utils::filter_empty which expects an array and consecutively throws a php warning Warning: array_filter() expects parameter 1 to be array, null given
Since the values are not set, the array is never created in the previous foreach (the foreach does not run) but it is accessed using the index value of 0 set here: https://github.com/CMB2/CMB2/compare/trunk...Sharsie:group-field-save-patch?expand=1#diff-fdb2b24e1079e11eb3bb6accda8a8f90R864

Not creating an issue for this as this is just a simple check before accessing a non existent array key

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -918,8 +918,9 @@ class CMB2 extends CMB2_Base {
 				$saved[ $field_group->index ][ $sub_id ] = $new_val;
 
 			}// End foreach().
-
-			$saved[ $field_group->index ] = CMB2_Utils::filter_empty( $saved[ $field_group->index ] );
+			if (isset( $saved[ $field_group->index ] ) {
+				$saved[ $field_group->index ] = CMB2_Utils::filter_empty( $saved[ $field_group->index ] );
+			}
 		}// End foreach().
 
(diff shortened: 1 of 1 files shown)


jtsternberg:
> When the repeatable group field is empty (all groups are removed) a null value is sent

When you remove the last group, it empties it, so there is always at least one group with empty fields, and the resulting value looks something like this:

```
$saved[ $field_group->index ]: Array
(
    [title] =>
)
```

Can you provide your box configuration you used which caused this issue? I can't imagine a scenario where `null` would be the value at that point in the execution.

jtsternberg:
Also, the tests are failing (missing a closing `)`: https://travis-ci.org/CMB2/CMB2/jobs/275383447), and the PR changes are not following the WordPress/CMB2 code standards.

tw2113:
No dimwits here in this thread @Sharsie. You're definitely not one, and i challenge anyone to change my mind. 

We appreciate the contributions here and hope to see continued contributions in the future :)

PR #1040, opened 2017-10-19
Fix issue with Field datetime_timezone storing the incorrect time stamp

jrmd:
Fixes an issue with the DateTime_timestamp_timezone fields where it would store an incorrect datetime object in the db

-  Alter how the field is sanitized and rendered to allow for the correct DateTime object to be stored and avoid any processing once the meta is fetched


tw2113:
What risks are present here with backwards compatibility? Are we going to risk breaking existing installs that use this field with the changes above?

jrmd:
Hmm, it could be; but as of now, the DateTime object isn't correct and does not match the epoch time like it should.

Within the current version of the plugin, when it creates the DateTime object it doesn't set the timezone until after the object creation which automatically works out the time difference. 

Example:
I set a date and time of `26/10/2017 12:00pm`, timezone `New York`.

Within the database it is stored like so: 
```s:128:"O:8:"DateTime":3:{s:4:"date";s:26:"2017-10-26 08:00:00.000000";s:13:"timezone_type";i:3;s:8:"timezone";s:16:"America/New_York";}";```

Which when unserialized has the incorrect time. Although, the _utc_ variant is the correct value when stored: 
```1509033600```

Since the timestamp should never change regardless of the timezone you shouldn't have to work out an offset. As the timestamp is just the epoch timestamp.

--- 

By introducing my fix, users will start to seeing accurate date and time in the panel to what is stored and if users resorted to using the UTC field instead it will not affect anything. This issue was apparently fixed in #227 but it doesn't seem to work anymore.

**Comment by jrmd** [comment]
@tw2113 Is there any update on this, I feel like this is quite a big issue with this field :)

**Comment by tw2113** [comment]
I always leave things up to @jtsternberg and his final say.

**Comment by jrmd** [comment]
@tw2113 Ahh okay thank you :)

**Comment by jonathanstegall** [comment]
I'm curious if this is still pending, or if it should have been resolved by something else? I seem to be having this issue as well.

**Comment by tw2113** [comment]
Safe to assume that if it's not merged in, it's not part of the official release or pending release yet. Apply to your own copy as necessary, helping confirm the issue being fixed with the proposed changes does help the likelihood of merging.

**Comment by jtsternberg** [comment]
The phpunit tests are failing on this. Can you investigate?

**Comment by jtsternberg** [comment]
Instructions for getting tests running: https://github.com/CMB2/CMB2/blob/develop/CONTRIBUTING.md#creatingrunning-tests

File: includes/types/CMB2_Type_Text_Datetime_Timestamp_Timezone.php [hunk 1]
```diff
@@ -34,6 +34,15 @@ public function render() {
 			$tz       = $datetime->getTimezone();
 			$tzstring = $tz->getName();
 			$value    = $datetime->getTimestamp();
```
toddsantoro, line 36: I fixed this issue of not displaying all dates and time correctly after saved by changing line 41 to:

$value = gmdate('m/d/Y h:i A', $datetime->getTimestamp());

Now all my dates show correctly and are still saved in the DB correctly: https://capture.dropbox.com/9QH9bxgAuUTPgVGt?src=ss rather than like this: https://capture.dropbox.com/Zl6aPMzCCo5rx4MY?src=ss

zodit: Also "text_datetime_timestamp" fields have the same bug, any chance to fix also CMB2_Type_Text_Datetime_Timestamp.php 
I am trying to apply your fix to the value, but it seems not working, could you help me please?

jtsternberg: Can you test the changes just pushed to the develop branch to confirm the issue is fixed for your use-cases?

Issue #712, opened 2016-08-17
issues with repeatable fields in a group and with sorting the groups


manmar:
I was using CMB2 v2.1.2 with WP v4.5.3 and had an issue with text repeatable fields within a group when sorting - the number of the repeatable fields would not move up or down when clicking on arrows. For example if the first item had 2 repeatable text fields and the second item had 3 text fields, after changing the order the second (now first) item would have only 2 text fields and the first would have 3 fields (one empty).
I've made a video to show what was going on exactly: http://recordit.co/COtIqkm5Mf

Then I updated to v2.2.2.1 (also tried 2.2.3.beta) and the same issue is still there but there is also a new one. Upon updating the page, two more groups are added. Next to last group will get the last repeating field value, and the last one will be empty.
Again, here is a video demonstrating the issue: http://recordit.co/Z8VgHU80hT

tw2113:
Can you provide your metabox configuration for this please?

It'd be everything with $something = new cmb2_meta_box() and the $something->add_field() bits.

Wanting to see if I can recreate the issue.

manmar:
Sorry for the wait, here is the configuration:

~~~~
add_action( 'cmb2_admin_init', 'mm_newsletter_archive' );

function mm_newsletter_archive() {

	$prefix = '_nla_';

	$nla = new_cmb2_box( array(
		'id' => $prefix . 'pdf',
		'title' => __( 'Newsletter Archive', 'cmb2' ),
		'object_types' => array( 'page' ),
		'context' => 'normal',
		'priority' => 'high',
		'show_on' => array( 'key' => 'page-template', 'value' => 'page-newsletter-archive.php' ),
	) );


	$nla_group = $nla->add_field( array(
		'id' => $prefix . 'group',
		'type' => 'group',
		'description' => __( 'Create a list of newsletters', 'cmb2' ),
		'options' => array(
			'group_title' => __( 'Newsletter {#}', 'cmb2' ),
			'add_button' => __( 'Add Another Newsleter', 'cmb2' ),
			'remove_button' => __( 'Remove Newsletter', 'cmb2' ),
			'sortable' => true,
		),
	) );


	$nla->add_group_field( $nla_group, array(
		'name' => __( 'Newsletter Title', 'cmb2' ),
		'id' => 'title',
		'type' => 'text',
	) );


	$nla->add_group_field( $nla_group, array(
		'name' => __( 'Newsletter URL', 'cmb2' ),
[...]

PR #1047, opened 2017-11-12
Fix repeating fields within repeating groups by changing regex depending on context

### LINKED ISSUE #1035: Repeatable custom field inside repeatable group don't save properly
I created custom field based on the address field
### Expected Behavior:
- The data need to be saved properly, each field created need to be saved.

### Actual Behavior:
- When adding repeatable custom field, say 3 for example, after save I only find 2.
- If there are for example 3 fields, and I added 4 other fields, after save, I only find 4 where the total should be 7
- Sometimes it will save all fields except the last one
- Saving doesn't work properly in general.

1. Create New Post
2. Scroll to Book details metabox, 
3. Create 3 fields (or any other number), give them random names
4. Update post (you'll see that only first 2 fields were saved)
5. Now add another 3 fields and update (you'll see that only the first one of them was saved)
6. If you add another group it will act the same way

I suspect it's javascript bug.

### CMB2 Field Registration Code:

https://gist.github.com/bomsn/36174de42bfbaa74abcd5bce1bfc432d

daggerhart:
Fixes #1035 and #348. Possibly related to #712 and others.

-  Pass the `group` boolean into `cmb.elReplacements` so it can know which index to replace for new Name and new IDs.

    -  If the new replacements are for a group, use a regex that won't match a field's attributes.
    -  If the new replacements are not for a group, use a regex that finds matches at the end of the field attribute strings, avoiding the group-level values within those attributes.

- When a new group is added, fix the name and ID attributes on the hidden `.empty-row`s so that the next "Add Row" has the correct attributes.

### DIFF
--- a/js/cmb2.js
+++ b/js/cmb2.js
@@ -421,6 +421,18 @@ window.CMB2 = window.CMB2 || {};
 					}
 				});
 			}
+
+			// Fix hidden empty rows
+			$row.find('.empty-row').find(cmb.repeatUpdate).each(function() {
+				var $emptyField = $( this );
+				var oldIndex = $emptyField.attr('id').split('_').pop();
+				var oldNameRegex = new RegExp('\\['+oldIndex+'\\]$');
+				var oldIdRegex = new RegExp('_'+oldIndex+'$');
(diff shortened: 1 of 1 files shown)


jtsternberg:
Nice, thanks!

PR #1051, opened 2017-11-17
 add description under options page title

modemlooper:
Fixes #{issue-number}.

-  

-  

-

### DIFF
--- a/includes/CMB2_Options_Hookup.php
+++ b/includes/CMB2_Options_Hookup.php
@@ -186,6 +186,9 @@ class CMB2_Options_Hookup extends CMB2_hookup {
 			<?php if ( $this->cmb->prop( 'title' ) ) : ?>
 				<h2><?php echo wp_kses_post( $this->cmb->prop( 'title' ) ); ?></h2>
 			<?php endif; ?>
+			<?php if ( $this->cmb->prop( 'description' ) ) : ?>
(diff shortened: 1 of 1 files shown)


modemlooper:
Example of description under title

option page description

jtsternberg:
This is currently possible using the before form hooks:

```php
$cmb_options = new_cmb2_box( array(
	'id'           => 'groupsboxid',
	'title'        => 'Groups',
	'description'  => 'Option to customize your groups',
	'object_types' => array( 'options-page' ),
	'option_key'   => '_yourprefix_theme_options', // The option key and admin menu page slug.
) );

add_action( "cmb2_before_options-page_form_{$cmb_options->cmb_id}", function( $object_id, $cmb ) {
	if ( $cmb->prop( 'description' ) ) : ?>
		<p><?php echo wp_kses_post( $cmb->prop( 'description' ) ); ?></p>
	<?php endif;
}, 10, 2 );
```
Not sure there's any justification for adding this inline.

modemlooper:
ok, didn't realize there was a hook. However, it would be easier to add as arg.

jtsternberg:
If the output location is important (or anything about the markup), you can use the `'display_cb'` parameter to override the default form output. https://github.com/CMB2/CMB2-Snippet-Library/blob/master/options-and-settings-pages/custom-display-callback.php#L49-L66

The main point being that this can be accomplished multiple ways, and I'd rather not add another parameter that is applicable _only_ to options page boxes.

PR #1052, opened 2017-11-19
Make CMB2_Option class properties public

ryanduff:
This class has two filters that pass `$this` but both properties are currently marked protected making this argument otherwise useless in the filter. Suggesting making them public so that any callbacks can access the properties and make use of them.

### DIFF
--- a/includes/CMB2_Options.php
+++ b/includes/CMB2_Options.php
@@ -48,14 +48,14 @@ class CMB2_Option {
 	 *
 	 * @var array
 	 */
-	protected $options = array();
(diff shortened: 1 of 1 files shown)


jtsternberg:
Would you be up to converting this to make the properties accessible via `__get` instead? It reduces the chances of breaking things as the key won’t be inadvertently stomped or changed.

jtsternberg:
E.g. https://github.com/CMB2/CMB2/blob/trunk/includes/CMB2_Base.php#L477

Issue #1055, opened 2017-11-24
Merge CMB2 Attached Posts Field into core


jtsternberg:
https://github.com/CMB2/cmb2-attached-posts
The idea here is to:

- [ ] Clean up/improve the UI. Would love contribution/help from @anhskohbo 
- [ ] Add support for other object types, Users, Terms, Post-Types of course.

jtsternberg:
Copying @daggerhart who is interested in helping

anhskohbo:
So, where's we can start the work?

daggerhart:
I have a morning full or errands to run, and then plan to work on it this afternoon. 

@anhskohbo, maybe wait for me to get a simple version into CMB2, then work from my fork? Alternatively, you could start working on the UI by forking `cmb2-attached-post` plugin, then copy your changes into my fork at a later time?

Those are just some random ideas. I'm open to anything.

anhskohbo:
OK, I'll wait for your fork

jtsternberg:
You bring up a good point. I will create a new branch and stub out the merge, then you can have something to build off. I should have done that yesterday. :)

jtsternberg:
OK, so new branch is up and I quickly stubbed out what a merge would look like, though I haven't tested it _at all_: https://github.com/CMB2/CMB2/tree/associated-objects-field

Please do your PRs against that branch.

@daggerhart would LOVE some help with you with testing and getting it to work initially and any of the todos you want to tackle: https://github.com/CMB2/CMB2/blob/d632d759a655057403cf51da2070ac10ce6e753b/includes/types/CMB2_Type_Associated_Object.php#L5-L10

@anhskohbo Until it's actually working, as @daggerhart mentioned, it's probably best to work in the plugin first: https://github.com/CMB2/cmb2-attached-posts, and we can get the updates folded in.

jtsternberg:
Ok, did a bit more work.
- [X] Started a PR (#1056) for more visible tracking of changes
- [X] Updated the css selectors to properly style the field. @anhskohbo this means you can probably start playing w/ cleaning up the UI on this branch rather than the plugin (if you haven't started)
- [X] Updated the field type. The type is `associated_objects`.

Along with checklist above, still need to:
- [ ] Fix the JS -- currently not working.

To test, you can use the example config: https://github.com/CMB2/cmb2-attached-posts/blob/master/example-field-setup.php#L20-L56
Just replace instances of `custom_attached_posts` with `associated_objects`

[...]

PR #1057, opened 2017-11-25
Query work on new associated-objects field type

daggerhart:
First pass at making field flexible for many different types of objects.

See: #1055

### DIFF
--- a/includes/types/CMB2_Type_Associated_Objects.php
+++ b/includes/types/CMB2_Type_Associated_Objects.php
@@ -4,7 +4,6 @@
  *
  * @todo Make field type for all object types, to and from.
  * @todo Maybe remove dependence on super-globals?
- * @todo Remove output buffer and instead concatenate string.
  * @todo Replace WP core `find_posts_div()` markup and ajax completely.
  * @todo Unit tests for field.
  * @todo Add example to example-functions.php. See https://github.com/CMB2/cmb2-attached-posts/blob/master/example-field-setup.php
@@ -39,7 +38,7 @@ class CMB2_Type_Associated_Objects extends CMB2_Type_Text {
 	 *
 	 * @var array
 	 */
-	protected $post_type_labels = array();
+	protected $object_type_labels = array();
 
 	/**
 	 * The type of field
@@ -63,6 +62,13 @@ class CMB2_Type_Associated_Objects extends CMB2_Type_Text {
 	protected $query_object_type = 'post';
 
 	/**
+	 * The field instance's query handler.
+	 *
+	 * @var CMB2_Type_Associated_Objects_Query
+	 */
+	protected $query;
+
+	/**
 	 * Constructor
 	 *
 	 * @since 2.2.2
@@ -79,6 +85,11 @@ class CMB2_Type_Associated_Objects extends CMB2_Type_Text {
 		}
 	}
 
+	/**
+	 * @param array $args
+	 *
+	 * @return CMB2_Type_Base|string
+	 */
 	public function render( $args = array() ) {
 		if ( ! is_admin() ) {
 			// Will need custom styling!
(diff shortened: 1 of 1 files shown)


includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -120,11 +135,12 @@ public function render( $args = array() ) {
 
 		// Open our retrieved, or found posts, list
 		echo '<div class="retrieved-wrap column-wrap">';
-		echo '<h4 class="associated-objects-section">' . sprintf( __( 'Available %s', 'cmb' ), $this->post_type_labels ) . '</h4>';
+		echo '<h4 class="associated-objects-section">' . sprintf( __( 'Available %s', 'cmb' ), implode('/', $object_type_labels ) ) . '</h4>';
```
jtsternberg: It's silly, but let's be sure to keep the WP standards (e.g. `implode( '/'...`)

jtsternberg: Also, it seems that the `implode()` part should maybe happen inside the `get_all_object_type_labels` method?

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -163,7 +179,7 @@ public function render( $args = array() ) {
 
 		// Open our attached posts list
 		echo '<div class="attached-wrap column-wrap">';
-		echo '<h4 class="associated-objects-section">' . sprintf( __( 'Attached %s', 'cmb' ), $this->post_type_labels ) . '</h4>';
+		echo '<h4 class="associated-objects-section">' . sprintf( __( 'Attached %s', 'cmb' ), implode('/', $object_type_labels ) ) . '</h4>';
```
jtsternberg: Same as above.

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -91,22 +104,24 @@ public function render( $args = array() ) {
...
+		// Set .has_thumbnail
+		$has_thumbnail = $this->field->options( 'show_thumbnails' ) ? ' has-thumbnails' : '';
+		$hide_selected = $this->field->options( 'hide_selected' ) ? ' hide-selected' : '';
+
 
 		// Check to see if we have any meta values saved yet
 		$attached = $this->field->escaped_value();
 		$attached = empty( $attached ) ? array() : (array) $attached;
 
+		$query = $this->get_query( $this->query_object_type );
```
jtsternberg: This is called 3 times. Is that necessary? Or can we just do `$this->query = $this->get_query(`... and use that throughout?

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -451,42 +323,44 @@ public function get_object( $id ) {
...
+		switch ( $query_object_type ) {
+			case 'user';
+				$query = new CMB2_Type_Associated_Objects_Query_Users( $this->query_args );
+				break;
+			case 'term':
+				$query = new CMB2_Type_Associated_Objects_Query_Terms( $this->query_args );
+				break;
+			case 'post':
+			default:
+				$query = new CMB2_Type_Associated_Objects_Query_Posts( $this->query_args );
```
jtsternberg: This class name is clean, but would maybe prefer a pattern that was a bit more concise. e.g. `CMB2_Type_Associated_Posts_Query`, leaving the base as `CMB2_Type_Associated_Objects_Query`.

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -607,3 +481,479 @@ public static function maybe_callback( $query, $post_args ) {
 	}
 
 }
+
```
jtsternberg: Once the other things are cleaned up, i'll move these to their own files, and probably rename them. Just a heads up.

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -607,3 +481,479 @@ public static function maybe_callback( $query, $post_args ) {
...
+		return $object->ID;
+	}
+
+	/**
+	 * @param WP_User $object
+	 *
+	 * @return string
+	 */
+	function get_object_type_label( $object ) {
+		return __('User');
```
jtsternberg: WP standards w/ () spacing.

includes/types/CMB2_Type_Associated_Objects.php
```diff
@@ -607,3 +481,479 @@ public static function maybe_callback( $query, $post_args ) {
...
+		return get_edit_term_link( $object->term_id, $object->taxonomy );
+	}
+
+	/**
+	 * @param $id
+	 *
+	 * @return WP_Term object
+	 */
+	function get_object( $id ) {
+		$found = NULL;
```
jtsternberg: Our coding style is lowercase on keywords.

PR #1082, opened 2018-01-12
Page/Screen specific hook

manzoorwanijk:
Using the Page/Screen specific hooks is the most efficient way to target a callback for only the screen(s) you need it for, instead of more generic `init`, `admin_init`, `admin_footer` etc.

Some of the page specific hooks are:

- `load-{page_hook}`,
- `admin_print_styles-{page_hook}` (used just before this hook)
- `admin_print_scripts-{page_hook}`
- `admin_head-{page_hook}`
- `admin_footer-{page_hook}`

So, the added hook will make it easy to make use of these hooks explicitly. One of the best uses of this hook can be using `load-{page_hook}` to call [add_screen_option](https://developer.wordpress.org/reference/functions/add_screen_option/).

### DIFF
--- a/includes/CMB2_Options_Hookup.php
+++ b/includes/CMB2_Options_Hookup.php
@@ -115,6 +115,13 @@ class CMB2_Options_Hookup extends CMB2_hookup {
 		 */
 		do_action( 'cmb2_options_page_hook', $page_hook, $this->cmb );
 
+		/**
+		 * Fires for specific $page_hook value
+		 *
+		 * @param CMB2 $cmb The CMB2 object
+		 */
+		do_action( 'cmb2_options_page_hook_' . $page_hook, $this->cmb );
+		
 		$this->maybe_register_message();
 	}
 
--- a/includes/CMB2_Options_Hookup.php
+++ b/includes/CMB2_Options_Hookup.php
@@ -106,6 +106,15 @@ class CMB2_Options_Hookup extends CMB2_hookup {
 			add_action( "admin_print_styles-{$page_hook}", array( 'CMB2_hookup', 'enqueue_cmb_css' ) );
 		}
 
+		/**
+		 * For other page/screen specific hooks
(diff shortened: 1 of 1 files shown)


tw2113:
No issue in my mind with merging this in. One thought I do have is perhaps add a second hook like so:

```
do_action( 'cmb2_options_page_hook_' . $page_hook, $this->cmb );
```
That way, the action hook itself would be specific to the $page_hook value, and no need for conditional checking. I know we do this elsewhere in CMB2. Provide a generic one for all, and provide dynamic ones as well, specific to the field type being run.

manzoorwanijk:
@tw2113 yes, adding one more for specific $page_hook would be a good option. Going to add and push that :)

jtsternberg:
I'm trying to understand why you would need this? Can you please provide example code as to how/why this would be important to you? Is the main reason so that you are able to gain access to the `$page_hook` variable?

manzoorwanijk:
@jtsternberg You may read the documentation of "load-{page_hook}" [here](https://github.com/WordPress/WordPress/blob/4.9-branch/wp-admin/admin.php#L214). It is explained how this hook can be used.
Let me give you an example. I'm a developing a plugin in which I need to make use of `WP_List_Table` and thus I need to provide some Screen Options. I would prefer to hook into the option page specific hook for `add_screen_option` instead of `admin_init` which is fired for every screen that I don't need to care about. Like this
```php
add_action( "load-{$page_hook}", 'add_custom_screen_option' );
function 'add_custom_screen_option() {
    $option = 'per_page';
    $args = [
        'label'   => 'Players',
        'default' => 5,
        'option'  => 'players_per_page'
    ];
    add_screen_option( $option, $args );
}
```

jtsternberg:
Going to close for now. I don't see a compelling reason to have this included. You can already hook in to this same spot in various ways. If you can provide a compelling use-case, I will reconsider. thank you.

PR #1088, opened 2018-02-09
Call remove_default_tax_metaboxes from any type of post_hooks

lipemat:
**v2.3.0** as well as tested against trunk.

**Issue**
Taxonomy fields accept a 'remove_default' parameter which unregisters the default term meta box on the post edit screens. This parameter works fine if you are using a standard meta box context when registering the meta box. However if you use another context such as 'after_title' this 'remove_default' parameter does nothing.

**To Recreate:**
Use a context when registering a box such as 'after_title'
Add a field of a taxonomy type and pass 'remove_default' to the field
See the default taxonomy meta box still exists on post edit screens

**Solution**
This pull request decouples the remove_default_tax_metabox() method from the add_metaboxes() method and calls it on it's own independent 'add_meta_boxes' action which allows the 'remove_default' param to work from any context.

### DIFF
--- a/includes/CMB2_hookup.php
+++ b/includes/CMB2_hookup.php
@@ -131,6 +131,7 @@ class CMB2_hookup extends CMB2_Hookup_Base {
 				add_action( 'add_meta_boxes', array( $this, 'add_metaboxes' ) );
 		}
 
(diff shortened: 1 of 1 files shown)


lipemat:
Closed due to realizing forked repo no longer exists. Created a new pull from the new repo.

PR #1098, opened 2018-02-19
Ability to autoload=false for options. Closes #1093

### LINKED ISSUE #1093: Enhancement: for options page - ability to autoload=false for options
### Expected Behavior:

When creating an options page using `\new_cmb2_box()` give ability to define the `autoloader`, so devs could define that those options should NOT be autoloaded.

As a developer, you know the drawbacks of `autoload=yes`.

### Actual Behavior:

currently it defaults to `yes`

Use the https://github.com/CMB2/CMB2-Snippet-Library/blob/master/options-and-settings-pages/options-pages-with-tabs-and-submenus.php snippet and save options.

jtsternberg:
Fixes #1093.

Ability to disable the options autoload parameter via filter (`"cmb2_should_autoload_{$options_key}"`) or via a box parameter for `'options-page'` box registrations (`'autoload' => false,`).

### DIFF
--- a/includes/CMB2_Options.php
+++ b/includes/CMB2_Options.php
@@ -193,7 +193,7 @@ class CMB2_Option {
 		$autoload = apply_filters( "cmb2_should_autoload_{$this->key}", true , $this );
 
 		// If no override, update the option
-		return update_option( $this->key, $this->options, $autoload ? null : false );
+		return update_option( $this->key, $this->options, (bool) $autoload );
 	}
 
 	/**
--- a/includes/CMB2_Options.php
+++ b/includes/CMB2_Options.php
@@ -180,8 +180,20 @@ class CMB2_Option {
 			return $test_save;
 		}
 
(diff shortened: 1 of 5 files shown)


includes/CMB2_Options.php
```diff
@@ -180,8 +180,20 @@ public function set( $options = array() ) {
...
+		 * @since 2.X.X
+		 *
+		 * @param bool        $autoload   Whether to load the option when WordPress starts up.
+		 * @param CMB2_Option $cmb_option This object.
+		 */
+		$autoload = apply_filters( "cmb2_should_autoload_{$this->key}", true , $this );
+
 		// If no override, update the option
-		return update_option( $this->key, $this->options );
+		return update_option( $this->key, $this->options, $autoload ? null : false );
```
slaFFik: @jtsternberg 
Here should be not `null` but explicit `true`.

PR #1108, opened 2018-03-14
Added 'id' attribute on group field

amans2k:
Hello,

Today one of my clients reported an error on his site.
'Cannot read property match of undefined ...'
src - https://i.imgur.com/1660qlC.png

While debugging I found, there was some code in client theme js file
which was trying to get 'id' of 'postbox' class element
src - https://i.imgur.com/xbjbTnX.png

and the page where the error was occurring has cmb2 group field.

I checked the cmb2 trunk and there was no id attribute
https://i.imgur.com/NaukHUd.png

I have added the 'id' attribute in the php and the js file as well.
run the grunt.

Kindly have a look.


jtsternberg:
I'm not sure I understand the implication that all CMB2 postbox elements should have IDs in order to compensate for a client theme js file?

tw2113:
I am inclined to agree with Justin here, and am wondering what's available for the theme to use that could more accurately find just what it's intending, or do some better checks for a found value before trying to do more.

amans2k:
I completely agree with you guys 

Themes should do proper checking before calling functions.

I have already given a patch code to the theme developers https://i.imgur.com/pTtIKIg.png

In WordPress, 'postbox' class is usually added to meta box wrapper https://i.imgur.com/TtFfaqe.png, and every meta box has 'id' attribute
And it is used in CMB2 group field as well without the 'id' attribute

This is a theme js conflict with CMB2 plugin. And there is a significant audience for both.

It's a small change in cmb2 that can prevent js error on their client sites.

jtsternberg:
I've merged this (the new working branch is develop). Thank you for your work here.

Issue #1118, opened 2018-04-19
GlotPress as the only translation solution


felipeelia:
Hi! First of all, congratulations for this excellent plugin!

As stated by @defries in #1067, plugins don't need to - and should not - use Transifex/WP Translations.org anymore. Nowadays GlotPress is the official way for translations, and a translation update should not be (1) a concern for plugin authors, (2) tied up to plugins versions.

The official page about this is [here](https://make.wordpress.org/meta/handbook/documentation/translations/). The third item says that the .mo/.po files shipped with the plugin are automatically imported (while translating to Brazilian Portuguese I could see that is true: some strigs were already there). As they were already imported, the plugin should not have those anymore. See the [WooCommerce folder](https://github.com/woocommerce/woocommerce/tree/master/i18n/languages), for instance.

Actually, the files generated by GlotPress override those bundled in plugin. For pt_BR, e.g., we can 100% sure remove the .po/.mo files (just made a test here).

That said, if you agree, we should:

- [ ] Replace the instructions about translations in readme files;
- [ ] Replace the instructions about translations in CONTIBUTING.md file;
- [ ] Remove .po/.mo files from languages dir.

Thanks!

jtsternberg:
@felipeelia I like your proposed solutions. Would you be able to provide a pull request addressing these issues?

felipeelia:
Sorry for the **very** late response here, @jtsternberg. I guess [this](https://make.wordpress.org/plugins/2016/03/01/please-do-not-submit-frameworks/) is the post @tw2113 was talking about, right?

I still think we should use the advantage of CMB2 being allowed in the WP repo and keep only the translations present in translate.wordpress.org. Keeping it in two different places doesn't help to keep consistency + everybody downloads all translations unnecessarily. It's possible to export the .po/.mo files from there (visit a [locale](https://translate.wordpress.org/projects/wp-plugins/cmb2/dev/pt-br/default/), scroll down and click on "Export"), so we'd have to update docs to properly point users to the needed steps to get the most recent translation in their languages.

Thinking about it a bit more thoroughly, is there any composer wizardry that could help devs who declare CMB2 as a project dependency? It'd be nice if when someone is fetching the plugin through composer it already gets the translation through WP export feature. It would require only two GET requisitions:
`https://translate.wordpress.org/projects/wp-plugins/cmb2/dev/pt-br/default/export-translations/`
and
[...]

PR #1119, opened 2018-04-19
Remove pt_BR translation files

felipeelia:
As explained in #1118, these files are not needed anymore :)

See https://translate.wordpress.org/locale/pt-br/default/wp-plugins/cmb2


jtsternberg:
The problem is that the translations still need to be included for those not using CMB2 as a plugin in the plugin directory.

felipeelia:
Yeah, I've made a comment about the current scenario [here](https://github.com/CMB2/CMB2/issues/1118#issuecomment-382816886).

jtsternberg:
Closing for now until the discussion on #1118 is resolved.

PR #1147, opened 2018-06-10
Call remove_default_tax_metaboxes from any type of post_hooks

lipemat:
**v2.3.0** as well as tested against trunk.

**Issue**
Taxonomy fields accept a 'remove_default' parameter which unregisters the default term meta box on the post edit screens. This parameter works fine if you are using a standard meta box context when registering the meta box. However if you use another context such as 'after_title' this 'remove_default' parameter does nothing.

**To Recreate:**
Use a context when registering a box such as 'after_title'
Add a field of a taxonomy type and pass 'remove_default' to the field
See the default taxonomy meta box still exists on post edit screens

**Solution**
This pull request decouples the remove_default_tax_metabox() method from the add_metaboxes() method and calls it on it's own independent 'add_meta_boxes' action which allows the 'remove_default' param to work from any context.

### DIFF
--- a/includes/CMB2_hookup.php
+++ b/includes/CMB2_hookup.php
@@ -131,6 +131,7 @@ class CMB2_hookup extends CMB2_Hookup_Base {
 				add_action( 'add_meta_boxes', array( $this, 'add_metaboxes' ) );
 		}
 
+		add_action( 'add_meta_boxes', array( $this, 'remove_default_tax_metaboxes' ) );
 		add_action( 'add_attachment', array( $this, 'save_post' ) );
 		add_action( 'edit_attachment', array( $this, 'save_post' ) );
 		add_action( 'save_post', array( $this, 'save_post' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


tw2113:
As a whole, I don't see anything drastically wrong with this. I am curious if the new add_action line would be better a few lines up in the switch statement section for the default case. The only other place I think it would be potentially needed is alongside `add_action( 'add_meta_boxes_comment', array( $this, 'add_metaboxes' ) );` but hard to say for sure.

Thoughts @lipemat ?

lipemat:
If we move it up into the switch, it won't work for any of the alternate contexts. Letting it run outside the switch ensures that it will run no matter which context is selected and in cases where there is only an alternate context used. 😄

tw2113:
The mentions being the different possible positions within the post editor screen alone, or post editor vs options page vs taxonomy term edit screen vs user profile? Just making sure I am comprehending correctly.

lipemat:
The different possible positions of the post editing screen. For instance if you add some fields to 'after_title' and want to remove the default taxonomy meta box (because you added a new field which contains a more appropriate way to set a term for your use case)

tw2113:
Just making sure.

jtsternberg:
Nice catch, nice fix. Thank you!

PR #1164, opened 2018-09-04
Updated the least required version

### LINKED ISSUE #1095: Fatal error in WordPress 3.8
Since it's mentioned in the plugin readme that WP 3.8 is supported, so I set the required version for my plugin to be 3.8. But, when I tested it on that version, it crashed.
**Fatal Error: Call to undefined function register_setting() in `/public_html/wp-content/plugins/cmb2/includes/CMB2_Options_Hookup.php on line 49`**

manzoorwanijk:
Fixes #1095 

The options page metaboxes crash for WP prior to 4.7 as discussed in #1095

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 Contributors:      jtsternberg, webdevstudios, tw2113
 Donate link:       https://cmb2.io
 Tags:              metaboxes, forms, fields, options, settings
-Requires at least: 3.8.0
+Requires at least: 4.7.0
 Tested up to:      4.9.6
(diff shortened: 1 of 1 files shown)


tw2113:
Thoughts on this one @jtsternberg ?

jtsternberg:
Since 4.6 isn't that long ago, I think I'd prefer to fix the issue. I'm thinking it might be as simple as including the right files before calling that function, but obviously will need testing.

manzoorwanijk:
Closing this in favor of #1166

PR #1166, opened 2018-09-04
Fix the options page issue in WP prior to 4.7

### LINKED ISSUE #1095: Fatal error in WordPress 3.8

manzoorwanijk:
Fixes #1095 and adds options page support for WP prior to 4.7, which otherwise crashed because of `register_setting` function been moved to `/wp-includes/option.php` in WP 4.7 because of [this](https://core.trac.wordpress.org/ticket/37885).

### DIFF
--- a/includes/CMB2_Options_Hookup.php
+++ b/includes/CMB2_Options_Hookup.php
@@ -297,7 +297,7 @@ class CMB2_Options_Hookup extends CMB2_hookup {
 			$url = add_query_arg( 'settings-updated', $updated ? 'true' : 'false', $url );
 		}
 
-		wp_safe_redirect( esc_url_raw( $url ), WP_Http::SEE_OTHER );
+		wp_safe_redirect( esc_url_raw( $url ), 303 );
 		exit;
 	}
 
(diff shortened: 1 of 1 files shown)


includes/CMB2_Options_Hookup.php
```diff
@@ -50,6 +50,17 @@ public function hooks() {
 			add_filter( "cmb2_should_autoload_{$this->option_key}", '__return_false' );
 		}
 
+		/**
+		 * for WP < 4.7
+		 *
+		 * @see https://core.trac.wordpress.org/ticket/37885
+		 */
+		global $wp_version;
+		if ( version_compare( $wp_version, '4.7', '<' ) ) {
```
jtsternberg: Let's use `if ( ! CMB2_Utils::wp_at_least( '4.7' ) && ! function_exists( 'register_setting' ) ) {`, but this looks good.

manzoorwanijk: yes, that's off course better :)

manzoorwanijk: Yes, I tested it for all: WP < 4.7, = 4.7 > 4.7.

jtsternberg:
Thanks again for your contribution. We very much appreciate it!

PR #1167, opened 2018-09-17
Add a check for file existence

manzoorwanijk:
Creating a custom field type whose classes (loaded by autoloader) start with CMB2 generated PHP warnings. So this conditional check prevents those warnings.

### DIFF
--- a/includes/helper-functions.php
+++ b/includes/helper-functions.php
@@ -41,7 +41,11 @@ function cmb2_autoload_classes( $class_name ) {
 		$path .= '/rest-api';
 	}
 
-	include_once( cmb2_dir( "$path/{$class_name}.php" ) );
+	$path = cmb2_dir( "$path/{$class_name}.php" );
+
+	if ( file_exists( $path ) ) {
+		include_once( $path );
+	}
 }
 
 /**


manzoorwanijk:
No, the issue is not like that. Let me explain.

If I create a custom field type and use autoloader. If my class names start with CMB2, then `cmb2_autoload_classes` is also called and that initial check is passed and thus a file which doesn't belong to CMB2 is included, thus giving a warning.

You can check by activating my custom field type [select_plus](https://github.com/manzoorwanijk/cmb2-select-plus).

tw2113:
Gotcha. I'm looking at the code and I do see the autoloading checks for "CMB2" in any part of the classname, so it isn't going to be a case of changing a small detail in your classname like I was first thinking.

@jtsternberg I see no issue with this check, and if you don't either, I propose we merge it in.

manzoorwanijk:
Yes, I can change the class name(s), but that check makes it to be safer and there should be no issues with that check.

jtsternberg:
@manzoorwanijk The main reason I would prefer to keep it as-is, is because it very loosely creates a requirement that 3rd-party developers should use their own prefixes, so as not to confuse the core-projects from them. Obviously, it's not possible to strictly enforce this, but I'm not keen to change the autloader to make it even easier to do.

manzoorwanijk:
@jtsternberg I understand the concern. Thanks :)

tw2113:
@manzoorwanijk can you try something for me, just for my own confirmation. Momentarily prefix your classes with something besides "CMB2_" and see if the issue remains. I believe the core autoloader may actually just be checking the first part of the name, which is not what I was thinking earlier.

manzoorwanijk:
@tw2113 I already did that. I named my class as `CMB2_Type_Select_Plus`, which passes that check by autoloader and thus causes the error.

PR #1168, opened 2018-09-17
Added new custom field type

manzoorwanijk:
Smart Select field for CMB2 with "optgroup" and "multiple" attribute support


tw2113:
Thanks.

Issue #1139, opened 2018-05-25
Tabbed navigation for options-page without submenus - with single option_key


manzoorwanijk:
As version 2.4.0 introduces tabbed navigation for options-pages. Is there a simple way to create those tabbed pages without creating the submenus? Preferably with single option_key

Example
![image](https://user-images.githubusercontent.com/18226415/40554826-9952822e-6064-11e8-9117-26925d7539c5.png)

manzoorwanijk:
@taunoha, that looks close enough but I wanted all the tab fields in single option key. The tab navigation should be JavaScript based without page reload.
I have now myself built something cool on those lines, just finishing it up :)

JiveDig:
I'm looking to do the same thing. Using the same option key, but breaking up the fields into tabs would be a great addition. I'll beed to roll something up myself as well I think.

taunoha:
Could you please explain why you need to use the same option key? What is the use case?

tw2113:
Using the same option key makes sense in my head at least, since it's still all one set of options, it's just visually separated out and organized via some tab sections.

manzoorwanijk:
@taunoha for example if there is a theme/plugin which has too many options to customise, it makes the page too long. It can easily be divided (visually into tabs). Creating separate menus and thus different option keys doesn't make sense in such case.

tw2113:
At least in regards to the submenu items getting added, I have whipped up this quick snippet which will feel really straightforward.

```
function wpcodex_adjust_the_wp_menu() {
	remove_submenu_page( 'MY-MENU-SLUG', 'MY-SUBMENU-SLUG' );
}
add_action( 'admin_menu', 'wpcodex_adjust_the_wp_menu', 999 );
```

Only downside thus far is that the Options page i'm doing this for loses menu highlight on the secondary tabs.

tw2113:
Perhaps a `show_submenu` parameter could be added for `options_page_menu_hooks()`

tw2113:
Any settings moved to new tabs will need to have option fallbacks for sake of fetching current settings for new setting.

Perhaps check for our current/new option first, and if that doesn't exist yet, check the original option name/value for a setting to use or save.

Also would be awesome to wire up the same idea above for when loading the new tab pages, and pre-populate. Utilize the filters in CMB2 for altering where to check for values.

pgroot91:
[...]

PR #1177, opened 2018-10-16
use "call_user_func()" instead of direct call.

manzoorwanijk:
Fixes the error caused when passing an object method to `display_cb`, therwise it results in **Fatal error**: **Function name must be a string in**


jtsternberg:
Hmm I don't have that issue. What version of php are you using, and can you provide your configuration where you are setting the `display_cb`?

manzoorwanijk:
> Hmm I don't have that issue. What version of php are you using, and can you provide your configuration where you are setting the `display_cb`?

I was testing my plugin on PHP 5.2 (still supported by WordPress) and faced that issue. I use a custom callback to display the settings page which has UI tabs as mentioned in #1139

jtsternberg:
Makes sense. I merged it and updated another instance as well. Thanks for the PR!

PR #1200, opened 2018-12-21
Remove superfluous method definitions

tnorthcutt:
This change removes unneeded method definitions from the class DocBlock. These methods are defined within the class, and so the `@method` tags aren't necessary (and in fact result in inspection errors in IDEs; in PhpStorm they trigger a "Method with same name already defined in this class" error).

Reference: http://docs.phpdoc.org/references/phpdoc/tags/method.html


jtsternberg:
👍

PR #1208, opened 2018-12-30
Add confirm dialog to group Remove button

### LINKED ISSUE #529: Add confirm dialog to Remove Group button
It worries me how easy it is to delete a group. Click and it's gone!

Can we get a confirm dialog in there?

At line 563 (in trunk today) make it like this for example:

   if ( number > 1 && confirm('Are you sure')) {

Ideally of course, that message should be customisable, and the confirm itself should be optional.

Thanks

slaFFik:
Fixes #529.

- added new option `remove_confirm` to the custom options list for `group` field type; `false` by default - so all the current installations will not see any difference, should be enabled to be used
- added `group_remove_confirm` - new translatable string to JS-based l10n list to display the test using JS
- display a default browser `confirm()` modal window and a custom (filterable) text from above when a user tries to delete a group of fields. If the user agrees - process further with the old logic, if denied - nothing happens. 

- [ ] Documentation should be updated [here](https://github.com/CMB2/CMB2/wiki/Field-Types#group).


js/cmb2.js
```diff
@@ -695,24 +695,30 @@ window.CMB2 = window.CMB2 || {};
 	cmb.removeGroupRow = function( evt ) {
 		evt.preventDefault();
 
-		var $this   = $( this );
-		var $table  = $id( $this.data('selector') );
-		var $parent = $this.parents('.cmb-repeatable-grouping');
-		var number  = $table.find('.cmb-repeatable-grouping').length;
+		var $this         = $( this ),
+			isConfirmable = $this.data('confirm') === 'yes';
```
jtsternberg: Please go back to the separate var statements-per-line as that is the code style of this lib. ([reasons](https://danhough.com/blog/single-var-pattern-rant/))

slaFFik: Older articles, but I see the point. Will update, thanks for the links!

slaFFik:
@jtsternberg Those changes were implemented.

I'm not sure what is a procedure in this project in this case (should I dismiss your initial review, or wait till you check your GH notifications for new commits etc) - so I'm just pinging you, just in case.

jtsternberg:
I just updated the docs.

PR #1210, opened 2019-01-01
Use phpunit 6.5 on PHP 5.6

slaFFik:
According to https://phpunit.de/getting-started/phpunit-4.html - phpunit 4 is for PHP 5.3-5.5.
phpunit 4-6 are for PHP 5.6.
Currently, on PHP 5.6 we install phpunit 4.8 which seems to just happen to work.
This PR adds a rule to install phpunit 6.5 on PHP 5.6.
Thus, we will use:
- special phpunit for PHP 5.2
- phpunit 4.8 for PHP 5.3-5.5
- phpunit 6.5 for PHP 5.6 and PHP 7.


slaFFik:
@jtsternberg Tests are failing because of #1209.

PR #1225, opened 2019-02-06
Don't exclude composer.json from the distribution

johnbillion:
## Description

When installing CMB2 via Composer, the composer.json file gets excluded from the package. While this file isn't strictly needed, it helps if this file is in place when you're checking dependencies in your project.

In addition, the composer.lock file _doesn't_ get excluded, which is weird.

## Motivation and Context

After installing CMB2 via Composer, I wanted to quickly check its dependencies but was unable to do so.

## Risk Level

Low

## Testing procedure

1. Run `mkdir composer-export && git archive HEAD | tar -xC composer-export`
2. Verify that composer.json is present in the `composer-export` directory

## Types of changes

## Checklist:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,7 +11,6 @@ node_modules       export-ignore
 package.json       export-ignore
 phpunit.xml        export-ignore
 tests              export-ignore
-composer.json      export-ignore
 readme.md          export-ignore
 README.md          export-ignore
 .idea              export-ignore


slaFFik:
`composer.lock` was added recently to the repo, and not added to `.gitattributes` to preserve it after `git archive` - that's why it's there, and `composer.json` is not.

jtsternberg:
I’d like to switch this as you mention and instead exclude the lock file. If you’re able to add that, great. If not, I can when I get to this.

jtsternberg:
Thanks!

joaoeduardo:
Remove composer.json from distribution breaks this composer plugin: `typisttech/imposter-plugin`

PR #1226, opened 2019-02-07
Ignore some more directories in the distribution package

johnbillion:
Related: #1225 and d77e89f.

This changes removes some more unnecessary directories from the package distribution.


jtsternberg:
Thanks!

PR #1231, opened 2019-03-04
Fixes #1230 — Removing multiple options fields reverts previous values

### LINKED ISSUE #1230: On options page preceding removed fields are undone
## Expected Behavior:
When clearing all fields (setting inputs to empty) on an options page and saving, the fields should be cleared and the option in the database should reflect this.

## Actual Behavior:
When an option is removed in CMB2 the last thing that happens is that the field_id is remove from the `CMB2_Options::$options_sets` for the given object_id [1].

For the next field, however, when the field is instantiated [2] the default is set for the given field which is retrieved. In this process, the `CMB2_Options::get()` method is called [3] to get the field's value. The problem is that if the corresponding `CMB2_Options::$option_sets` for the given object_id does not exist **or is `empty()`** then it returns a new CMB2_Option [4]. This means that if the preceding field removed its value then then `empty( self::$option_sets[ $option_key ] )` will return true. This causes it to retrieve the options again [5] and that causes it to overwrite the previously removed value as the removed field hasn't been stored in the database, yet.

[1] https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Field.php#L440
[2] https://github.com/CMB2/CMB2/blob/develop/includes/CMB2.php#L800
[3] https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Field.php#L292
[4] https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Field.php#L292
[5] https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Options.php#L223

## Possible Solution
I'm still working towards a solution. Clearly the issue is that the option is being retrieved again and overwriting the `CMB2_Options->options` property, which is later used to update the options. The trick is finding at what point in the chain something different should be happening.

### Possible Solution's Risk Level
No proposed solution, yet. But will try to find one.

## Steps to reproduce:

1. Create an options page using the fields below.
2. Fill both fields with some value and save.
3. Empty only the second field and save.
4. With the second field empty, try emptying the first field and saving. When the page reloads the first field's old value persists.

## CMB2 Field Registration Code:

```php
        $metabox = new \CMB2([
            'id'           => "test_options",
            'title'        => 'MessageWP Options',
            'object_types' => ['options-page'],
            'option_key'   => "test_options",
            'parent_slug'  => 'options-general.php',
            'menu_title'   => 'MessageWP',
        ]);

        $metabox->add_field([
            'desc'        => 'License key for plugin updates. Clear to deactivate license.',
            'name'        => 'License Key',
            'type'        => 'text',
            'id'          => "test_license"
        ]);

        $metabox->add_field([
            'name'            => 'Base URL',
            'type'            => 'text',
            'id'              => "test_slug",
            'attributes'      => [
                'placeholder' => 'messages'
            ],
            'sanitization_cb' => function ($value) {
                return empty($value) ? $value : sanitize_title($value);
            }
        ]);
```

## Your Environment
Browser name and version: Chrome v72
Operating System and version (desktop or mobile): macOS Mojave

JasonTheAdams:
This fixes #1230, the solution was proposed by @manzoorwanijk

### DIFF
--- a/includes/CMB2_Options.php
+++ b/includes/CMB2_Options.php
@@ -48,7 +48,7 @@ class CMB2_Option {
 	 *
 	 * @var array
 	 */
-	protected $options = array();
+	protected $options = null;
 
 	/**
 	 * Current option key
@@ -212,7 +212,7 @@ class CMB2_Option {
 	 * @return mixed          Value set for the option.
 	 */
 	public function get_options( $default = null ) {
-		if ( empty( $this->options ) && ! empty( $this->key ) ) {
+		if ( null === $this->options && ! empty( $this->key ) ) {
 
 			$test_get = apply_filters( "cmb2_override_option_get_{$this->key}", 'cmb2_no_override_option_get', $default, $this );
 


manzoorwanijk:
Before merging it, we need to make sure that the initial `null` value is not returned where an array may be expected, just not to break the existing code. For example see the `return` statement [here](https://github.com/CMB2/CMB2/blob/v2.6.0/includes/CMB2_Options.php#L80). May be we should test such cases.

JasonTheAdams:
Good call, @manzoorwanijk. Here's what I've found.

`CMB2_Option::delete_option()` isn't used anywhere in the entire CMB2 codebase.

`CMB2_Options::remove()` is used when removing field values and it appears to work properly when a null value is returned. It's ultimately stored in `$updated` [here](https://github.com/CMB2/CMB2/blob/v2.6.0/includes/CMB2_Field.php#L572). That may changed the following condition where `$updated` is checked, but this seems more accurate and I haven't experienced any bugs.

This is being tested in the current unit tests, most of which passed on Travis (the nightly and one of the PHP 7.1 runs failed, but it doesn't look like it was the actual unit test but was something else).

manzoorwanijk:
> `CMB2_Option::delete_option()` isn't used anywhere in the entire CMB2 codebase.

May be some plugin/theme uses it?

JasonTheAdams:
That's entirely possible, of course, but CMB2 can't really control whether other plugins/themes assume that the returned value is always an array. Fortunately, I believe in most instances that a delete method is being called, the return value is most often ignored. I'm open to push-back on this, though.

PR #1235, opened 2019-03-10
Add opening of postbox for HTML5 validation

### LINKED ISSUE #1233: required field within closed group
Required fields inside a closed (collapsed) group lead to a confusing UI experience.

When a required field is ignored, the UI normally pops up a prompt that the user should fill out the missing information, but if that field is in a closed group... clicking Publish/Update results in nothing. No response from page to indicate the problem.

The group should be forced open so that the UI can point the user to the problem.

aurmer:
## Description
Added one feature to cmb.init. It removes the "closed" class from each Postbox created by CMB2 iff that Postbox has a required input inside of it.
HTML5 validation (specifically 'required' input fields) throws JS error if the field is incomplete AND hidden. The browser (chrome) UI provides no feedback as to the cause of the Submit not functioning.

## Motivation and Context
This solves a UI problem. The current version relies on HTML5 form validation protocol to enforce "required" fields, but that protocol doesn't provide UI feedback when the "required" field is set to "display:none".
Fixes #1233

## Risk Level
admin-only = minimal risk
This only adds an event listener to "id=publish". If that element doesn't exist on the page, this feature does nothing.

## Testing procedure
Create a custom metabox containing required fields.
Add a group containing more required fields.
Add a group containing no required fields.
Open relevant edit page.
Leave required fields blank.
Close all groups and custom metabox.
Click Publish/Update.
Test passes if only the custom metabox and the first group are automatically opened. Second group shouldn't be affected because it doesn't have a required field.

## Types of changes
**New feature (non-breaking change which adds functionality)**
This ensures the user will get feedback in the event a required field is hidden in a closed group or custom metabox

## Checklist:
I am unsure about the "code style" of the project, but I did my best to match.
- [ ] My code follows the code style of this project.
- [x] My code and pull requests meets the [Contributing guidelines]


tw2113:
Some initial feedback.

Nitpick on the indent needing one more tab to get aligned.

The rest is that while I can appreciate using vanilla javascript for it all, we do have jQuery available to use in the entire spot, and have already cached the `document` in `$document` as an example.

I think we may process the js etc before final releases, so that may not be needing to be done.

Lastly, since we do have the minified versions as well that get processed and made, there's room to use more meaningful variable names than `x` in the developer copy. Helps make developing understand the code a bit better.

All in all, it looks like it's valid and working code, so i think it'll be good to merge in in the end, pending the feedback above and any discussion needed there.

Issue #1239, opened 2019-03-12
Change cmb2 label for fields group for rest api


Mte90:
## Expected Behavior:
When you open a rest api page you see the fields under the group of `cmb2` but this should be possible to change it based on the needs.

## Actual Behavior:
Right now is not possible.

## Possible Solution
Add a filter there https://github.com/CMB2/CMB2/blob/develop/includes/rest-api/CMB2_REST.php#L237

jtsternberg:
To change what? the fields that show?

Mte90:
![immagine](https://user-images.githubusercontent.com/403283/54206432-91b9ac00-44d8-11e9-9718-06f4bc2c18f2.png)
As you can see here cmb2 groups all the fields in that group but for a public endpoint is not beautiful and clear what means.
If it is possible to change will be better :-)

jtsternberg:
Ah! you mean change the `cmb2` prefix. That makes sense, but it's also problematic as it could break other plugins or theme which is depending on the response looking that way. This would need some thought.

Mte90:
I see that is not easy to do that change because the group name is used to check in the rest of that class to show the values.
Probably add that feature means rewrite all this class to add support for a big astraction (and avoid conflicts with other plugins).

Maybe we can add a value in the metabox that is used to define the group and change the flow of the various methods.

jtsternberg:
Yes, that could definitely work. We could use the [foreach loop](https://github.com/CMB2/CMB2/blob/associated-objects-field/includes/rest-api/CMB2_REST.php#L176) to group all the boxes into their defined rest slug. It will take a bit of work to change the logic here, but could definitely be done.

Mte90:
I am starting to experimenting with https://github.com/Mte90/CMB2/commit/e664f73ce780a8069c02632473a80bff139b36a9
It is not perfect because should ignore the metabox of other post types anyway and is not rewriting the flow.

Mte90:
With this other commit https://github.com/Mte90/CMB2/commit/125848c0e562620ebf5add566faf7170b030493b seems that now is altro backward compatible and don't conflicts with other post type or metabox.
I did that without rewrite everything as suggested so I am not sure if it is the case to do a pr for it.

How works:

* Load the metabox
* Check for a parameter in the metabox that is the slug for the api
* Owerwrite for that metabox instance the slug

In this way don't create conflicts and we don't need to refactor the class.

PR #1240, opened 2019-03-13
Prototype for #1239, change slug for rest fields group

Mte90:
## Description
immagine
This pr enable with a new parameter in the metabox to change the `cmb2` slug with something custom without creating conflicts with other mentabox in REST api.

## Motivation and Context
For transaprence and improve the code organization this improve the quality of the schema of CMB fields.

## Risk Level
This code shouldn't create issues with backward comaptibility because `cmb2` is a default value.

## Testing procedure
Create a new metabox, enable `show_in_rest` and also `rest_group_values` and define a slug.
Probably we need a better parameter that is more explicative then this one about the purpose.

## Checklist:

### DIFF
--- a/includes/rest-api/CMB2_REST.php
+++ b/includes/rest-api/CMB2_REST.php
@@ -72,7 +72,14 @@ class CMB2_REST extends CMB2_Hookup_Base {
 	 * @since 2.2.3
 	 */
 	protected $edit_fields = array();
-
+	
+	/**
+	 * The CMB2 object rest slug.
+	 *
+	 * @var boolean
+	 */
+	protected static $rest_slug = 'cmb2';
+	
 	/**
 	 * Whether CMB2 object is readable via the rest api.
 	 *
@@ -117,6 +124,9 @@ class CMB2_REST extends CMB2_Hookup_Base {
 	public function __construct( CMB2 $cmb ) {
 		$this->cmb = $cmb;
 		self::$boxes[ $cmb->cmb_id ] = $this;
+		if( !empty( $this->cmb->prop( 'rest_group_values' ) ) ) {
+            		self::$rest_slug = $this->cmb->prop( 'rest_group_values' );
+		}
 
 		$show_value = $this->cmb->prop( 'show_in_rest' );
 
@@ -234,7 +244,7 @@ class CMB2_REST extends CMB2_Hookup_Base {
 	 * @return void
 	 */
 	protected static function register_rest_field( $object_types, $object_type ) {
-		register_rest_field( $object_types, 'cmb2', array(
+		register_rest_field( $object_types, self::$rest_slug, array(
 			'get_callback'    => array( __CLASS__, "get_{$object_type}_rest_values" ),
 			'update_callback' => array( __CLASS__, "update_{$object_type}_rest_values" ),
 			'schema'          => null, // @todo add schema
@@ -319,7 +329,7 @@ class CMB2_REST extends CMB2_Hookup_Base {
 	 * @return mixed
 	 */
 	public static function get_post_rest_values( $object, $field_name, $request, $object_type ) {
-		if ( 'cmb2' === $field_name ) {
+		if ( self::$rest_slug === $field_name ) {
 			return self::get_rest_values( $object, $request, $object_type, 'post' );
 		}
 	}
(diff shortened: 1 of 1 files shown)


includes/rest-api/CMB2_REST.php
```diff
@@ -117,6 +124,10 @@ public static function maybe_init_and_hookup( CMB2 $cmb ) {
 	public function __construct( CMB2 $cmb ) {
 		$this->cmb = $cmb;
 		self::$boxes[ $cmb->cmb_id ] = $this;
+		self::$rest_slug = 'cmb2';
```
slaFFik: @Mte90 What's the point of this line (127) when we have a predefined value on line 81?

includes/rest-api/CMB2_REST.php
```diff
@@ -117,6 +124,10 @@ public static function maybe_init_and_hookup( CMB2 $cmb ) {
...
+		if( !empty( $this->cmb->prop( 'rest_slug_group' ) ) ) {
```
slaFFik: @Mte90 Please improve code style - there should be a space after `!`: `! empty()`.

includes/rest-api/CMB2_REST.php
```diff
@@ -117,6 +124,10 @@ public static function maybe_init_and_hookup( CMB2 $cmb ) {
...
+            		self::$rest_slug = $this->cmb->prop( 'rest_slug_group' );
```
slaFFik: @Mte90 Please don't mix tabs and spaces on the same line. Please switch to WordPress Coding Standards - using tabs only in PHP.

slaFFik: Also, I'm not sure that the `prop` method sanitizes the value. Please check that and if not - would be nice to have something like `sanitize_key()` here.

jtsternberg: Nah, we don't need to sanitize this. The reason is that CMB2 config is meant to be coded (not added via user input/db), so if we needed to sanitize this, we would need to sanitize a bunch of other stuff in CMB2 that we don't (like field names/ids, etc).

includes/rest-api/CMB2_REST.php
jtsternberg: Also, you can't use a an evaluated expression in `empty` in earlier versions of PHP. E.g. the value inside `empty()` needs to be a variable.

jtsternberg:
I'm going to leave this PR open, as it's described a prototype, but I won't merge as-is. The reason is that I don't want someone using the `rest_slug_group` group to inadvertently change the namespace for everyone else using CMB2. Instead we will need to put only the fields within that specific box in that new namespace, so will require a bit of structure change (https://github.com/CMB2/CMB2/issues/1239#issuecomment-472024661).

includes/rest-api/CMB2_REST.php
Mte90: I saw that in case of multiple metabox where the parameter is not defined without that line, in other rest pages where should be the default value was replaced.
With forcing a reset the code was working and keeping backward compatibility.

Mte90:
Yes it is only a prototype that works and not create issues with other metabox as it is right now.
I prefered to do a pr just in case someone need an implementation or to have something to start discussing.
I can work on improve the code style anyway, it is not a big problem :-)

jtsternberg:
I appreciate the effort here. Will look into testing this as soon as I can.

Mte90:
Well CMB2 is very important for my job since years so I am happy to contribute to it as much I can :-)

Mte90:
any updates for this pr?

jtsternberg:
This PR does not work. It probably works in your tests because you only have 1 or 2 metaboxes and test types, and registered your `rest_slug_group` in the right order. it doesn't work because the `register_rest_field` methods happen in a static context, and do load like you think (per CMB2 object). In order to fix this, it will need some significant infrastructure changes.

Mte90:
Quite sad but I hope for a fix that works in all the cases :-)

jtsternberg:
@Mte90 I just pushed a couple updates to the develop branch that will allow registering your own handlers for `register_rest_field`. This example replaces the rest slug, which you're trying to accomplish. 
https://gist.github.com/jtsternberg/a70e845aca44356b8fbf05aafff4d0c8

It's a bit more verbose than what you were trying to do, but it does work, and provides ultimate flexibility.

Mte90:
I hope that will be documented but thanks a lot!

PR #1251, opened 2019-04-10
Feature: Adding possibility to use the meta value of a group field row in the collapsed heading

Blackbam:
Feature: Adding possibility to use the meta value of a group field row in the collapsed heading

## Description
I was always missing this feature because if you have like 200 collapsed rows in a group field repeater you may search for hours for the correct entry. With this easy extension it is possible to add values from the group field to the title of the row. This is done by {field_id} (see example below).

## Motivation and Context
Not sure if an issue for this exists but I think it is an important new feature which I need for various projects.

## Risk Level
Minimal risk - as long as you do not put {meta_name} into the group fields title nothing can go wrong.

## Testing procedure
Look at the following example:

```
 $blog_group_id = $cmb->add_field( array(
        'id'          => 'endpoints',
        'type'        => 'group',
        'repeatable'  => true,
        'options'     => array(
            'group_title'   => __('Endpoint {#}: {path} ',CERR_PLUGIN_TEXTDOMAIN),
            'add_button'    => __( 'Add new', CERR_PLUGIN_TEXTDOMAIN ),
            'remove_button' => __( 'Remove', CERR_PLUGIN_TEXTDOMAIN ),
            'closed'        => true,  // Repeater fields closed by default - neat & compact.
            'sortable'      => true,  // Allow changing the order of repeated groups.
        ),
    ) );`

    $cmb->add_group_field( $blog_group_id, array(
        'name' => 'Path',
        'id'   => 'path',
        'type' => 'text',
    ) );
```

This extension only affects the "group_title" within the output. Endpoint {#}: {path} for instance may translate to Endpoint 1: /foo/bar

Please consider adding this to the core it would be valuable in my opinion. In case I have done anything wrong please tell me how I shall improve it.

### DIFF
--- a/includes/CMB2.php
+++ b/includes/CMB2.php
@@ -592,6 +592,21 @@ class CMB2 extends CMB2_Base {
 		$confirm_deletion = $field_group->options( 'remove_confirm' );
 		$confirm_deletion = ! empty( $confirm_deletion ) ? $confirm_deletion : '';
 
+        // for replacing {my_field_id} with the corresponding field value in repeaters
+        $replace_with_meta = function($heading) use ($field_group) {
+            if(isset($field_group->value[$field_group->index])) {
+                $values = $field_group->value[$field_group->index];
+                $fields = $field_group->args('fields');
+                foreach($fields as $id => $properties) {
+                    if(strpos($heading, '{'.$id.'}') !== false) {
+                        $replace = (is_scalar($values[$id])) ? strval($values[$id]) : '';
+                        $heading = str_replace('{'.$id.'}',$replace,$heading);
+                    }
+                }
+            }
+            return $heading;
+        };
+
 		echo '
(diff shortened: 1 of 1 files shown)


Blackbam:
Just noted that there is a little problem with the JavaScript when adding new lines which maybe should be fixed.

slaFFik:
CMB2 plugin currently supports PHP 5.2. 
Your PR will break all sites on this version as it used closures.
Also, all the code style should follow [WordPress PHP Coding Standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/).

Blackbam:
Ok did not know that you still support PHP 5.2. I will implement with legacy support and I will also try to respect the WP PHP coding style.

tw2113:
@Blackbam @slaFFik Remember that WordPress core drops PHP 5.2 support in WordPress 5.2, and we'll be able to use 5.6+. Not sure how soon we'll be trying to require WP5.2 but it's an available detail

Issue #1260, opened 2019-04-14
VIP feedback


jtsternberg:
@GaryJones has provided some feedback to get CMB2 in a better state for use on VIP sites. I'm going to list his feedback here.

> Hey - here’s the first real batch of CMB2 issues I’ve found (basically from one file). See what you think.

- includes/CMB2:365 - should be `sanitize_html_class`, and then wrap in `esc_attr()`.
- includes/CMB2:511 - needs `esc_html()` around `$label`
- includes/CMB2:514 - needs `esc_html()` around `$description`
- includes/CMB2:529 - needs `esc_html()` or `wp_kses_*()` around button content
- includes/CMB2:574 - potentially incorrect use of `sanitize_html_class()`
- includes/CMB2:574 - no need for `strip_tags`, as `sanitize_text_field()` does that already.
- includes/CMB2:596 - `id` attribute value needs esc_attr()
- includes/CMB2:596 - `data-iterator` attribute value needs `esc_attr()`
- includes/CMB2:599 - `data-selector` attribute value needs `esc_attr()`
- includes/CMB2:604 - h3 content needs `esc_html()`
- includes/CMB2:627 - `data-selector` attribute value needs `esc_attr()`
- includes/CMB2:627 - `data-selector` attribute value needs `esc_attr()`
- includes/CMB2:627 - needs `esc_html()` or `wp_kses_*()` around button content
- includes/CMB2:1011 - `$_REQUEST` input needs sanitization - `absint()`, `sanitize_*()`, etc.
- includes/CMB2:1016 - `$_REQUEST` input needs sanitization - `absint()`, `sanitize_*()`, etc.
- includes/CMB2:1021 - `$_REQUEST` input needs sanitization - `absint()`, `sanitize_*()`, etc.
- includes/CMB2:1033 - `$_REQUEST` input needs sanitization - `absint()`, `sanitize_*()`, etc.
- includes/CMB2:1199 - `$_GET['page']` input needs sanitization - `sanitize_text_field()`, etc. though somewhat mitigated by `options_page_keys()`
- includes/CMB2:1203 - `$_POST['action']` input needs sanitization - `sanitize_text_field()`, etc.  though somewhat mitigated by `options_page_keys()`
- includes/CMB2:1206 - The above two items could be mitigated if only a boolean is returned.
[...]

PR #1262, opened 2019-04-19
Add late escaping to CMB2_Display_File

mikeselander:
## Description

Picking off some escaping/sanitizing from #1260. This PR late-escapes output from the `file_output` method of the `CMB2_Display_File` class. I realize that we're already escaping the URL in https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Field_Display.php#L397, but late-escaping is preferred and we could possibly get rid of the early-escaping from the link above.

## Motivation and Context
Hardening escaping and sanitizing throughout the plugin.

## Risk Level
Minimal risk

## Types of changes

## Checklist:


jtsternberg:
👍

PR #1271, opened 2019-05-23
Add function_exists('add_action') check

### LINKED ISSUE #1270: Add function_exists( 'add_action' ) check to initialization file
When using CMB2 via composer and using a PHP based tool (e.g. PHP Code Sniffer), the PHP based tool throws an error.

## Expected Behavior:

When installing CMB2 and PHP Code Sniffer via composer with a `composer.json` of

```
{
    "require": {
        "cmb2/cmb2": "^2.6"
    },
    "require-dev": {
        "squizlabs/php_codesniffer": "3.*"
    },
    "autoload"    : {
        "files": [
            "vendor/cmb2/cmb2/init.php"
        ]
    },
    "extra": {
        "installer-paths": {
            "vendor/cmb2/cmb2": ["cmb2/cmb2"]
        }
    }
}
```

I expect the following to run PHP Code Sniffer (`phpcs`)

```
$ composer install
$ ./vendor/bin/phpcs
```

## Actual Behavior:

I get a fatal error because the function `add_action()` is undefined.

```
$ composer install
$ ./vendor/bin/phpcs
Fatal error: Uncaught Error: Call to undefined function add_action() in vendor/cmb2/cmb2/init.php:126
```

## Possible Solution

If we add a `function_exists()` check for `add_action()` before calling it, we can prevent throwing this fatal error.

### Possible Solution's Risk Level

This seems to be a relatively low risk change. Since we're checking if the function `add_action()` exists directly before calling it, the only behavior change we should get is avoiding throwing this fatal error.

## Steps to reproduce (I have confirmed I can reproduce this issue on the [`develop`](https://github.com/CMB2/CMB2/tree/develop) branch):

1. Clone https://github.com/salcode/salcode-cmb2-autoload-demo
2. Run `composer install`
3. Run `./vendor/bin/phpcs`

## Your Environment

Mac OS X Command Line

salcode:
Add function_exists('add_action') check to initialization file. When
loading CMB2 via Composer (and autoloading init.php), the init.php will
be loaded every time. Usually this is not a problem, however if we are
also using Composer to load a PHP development tool (e.g. PHP Codesniffer
(phpcs)), when the development tool runs it is NOT in the context of
WordPress.

Running CMB2's init.php outside of the context of WordPress
results in a fatal error because the function add_action() does not
exist. By adding this function_exists() check, we can avoid this fatal
error when autoloading init.php outside of the context of WordPress.

See #1270

## Description

Check that the `add_action()` function exists before calling it in `init.php`.

## Motivation and Context

Fixes #1270.

## Risk Level
This seems to be a relatively low risk change. Since we're checking if the function `add_action()` exists directly before calling it, the only behavior change we should get is avoiding throwing this fatal error.

### develop tests are failing

The tests on the `develop` branch (before adding any work of my own) are currently failing. 

```
FAILURES!
Tests: 210, Assertions: 690, Failures: 6.
```

### Adding my changes to master

I cherrypicked the one commit in this PR onto the `master` branch and ran `phpunit` there.

All tests passed when my change was introduced into the `master` branch.

```
...............................................................  63 / 210 ( 30%)
............................................................... 126 / 210 ( 60%)
............................................................... 189 / 210 ( 90%)
.....................                                           210 / 210 (100%)

Time: 9.44 seconds, Memory: 44.00MB

OK (210 tests, 690 assertions)
```

## Types of changes

## Checklist:

### DIFF
--- a/init.php
+++ b/init.php
@@ -123,6 +123,11 @@ if ( ! class_exists( 'CMB2_Bootstrap_260_Develop', false ) ) {
 				define( 'CMB2_LOADED', self::PRIORITY );
 			}
 
+			if ( ! function_exists( 'add_action' ) ) {
+				// We are running outside of the context of WordPress.
+				return;
+			}
+
 			add_action( 'init', array( $this, 'include_cmb' ), self::PRIORITY );
 		}
 


jtsternberg:
Seems legit. Thank you @salcode

johnbillion:
I don't believe this change makes much sense. Using Composer to autoload a WordPress plugin with side effects is incorrect and shouldn't be fixed by coding around the side effects in every plugin.

See my comment on the same issue reported against the action-scheduler for my reasoning: https://github.com/Prospress/action-scheduler/issues/303#issuecomment-500409458

PR #1274, opened 2019-06-03
Encode value to avoid php warning of converting array to string

quangthien27:
## Description
Check if value is an array before formatting the string

## Motivation and Context
Avoid PHP warning

## Risk Level
admin-only


jtsternberg:
Thank you for your submission, however I will not merge it as it is unnecessary. If you want json-encoded attributes, you should json-encode them yourself when you set the attributes. There are a couple examples commented out in the example-functions.php file:

https://github.com/CMB2/CMB2/blob/479ad072381c51e265fcbf070b1259ab6467b581/example-functions.php#L284-L288

and
https://github.com/CMB2/CMB2/blob/479ad072381c51e265fcbf070b1259ab6467b581/example-functions.php#L310-L318

PR #1276, opened 2019-06-05
Character counter

### LINKED ISSUE #1241: Character counter / max option
A project of ours needs a character counter, applicable to WYSIWYGs and textareas. It would show how many characters have been entered, and optionally enforce a limit.

Just wanted to check (1) if you've anything in the pipeline along these lines, and (2) if not, would it be something you'd be interested in as a contribution to CMB2, or would you prefer a separate plugin?

I'm thinking of 2 new field parameters, applicable to textareas and WYSIWYGs:

- `char_counter` (true / false)
- `char_max` (integer, only works if `char_counter` is true)

gyrus:
Addition of field parameters which provide functionality for character counters, and character limit enforcement.

## Description

- Functionality for text, textarea  and WYSIWYG field types only.
- Not available for grouped WYSIWYG fields.

## New field parameters

- `char_counter`: `'characters'` | `'words'`
- `char_max`: integer. When defined, counter shows remaining chars / words.
- `char_max_msg`: string, default: `'Your text may be truncated.'`
- `char_max_enforce`: boolean, default: `false`. Currently only applied (as `maxlength` attribute) to text and textarea fields which use 'characters' for counter.

## Motivation and Context

This was a suggested feature which I needed for a project and seemed like a useful addition to the plugin.

Fixes #1241 

## Risk Level

Medium risk? Has been tested on front-end forms.

## Testing procedure

- Tested in a local vanilla WP installation with Twenty-Fifteen theme and no other plugins (admin only).
- Also tested on a live client project with a complex custom theme and 40+ other plugins - both admin and front-end forms.
- Tested in latest Chrome, Safari, Firefox (Mac), IE11 (Win10 / Win7) and MS Edge (Win10).
- The only issue I've found is an 'Uncaught TypeError' in console when using the Chrome Form Filler extension to fill a field with dummy content. However, functionality doesn't seem to be affected.

## Types of changes

## Checklist

- Unfortunately I'm not familiar enough with unit testing to have written unit tests (or be sure they're even needed!).
- I can prepare all proper docs for CSS classes, etc.

## IMPORTANT NOTES

- Due to problems with jshint, `cmb2.min.js` isn't currently working. JS all works fine when unminified scripts are used in a dev environment. AFAIK the problem is with the use of `_.debounce()` in `cmb2-char-counter.js`. This is copied from WP core handling of TinyMCE. The `_` is available in core. Either Grunt jshint needs modifying so it can recognise core code as being available, or I need an alternative debounce.
- In any case, the code here needs a proper build process to be run to get `cmb2.min.js` properly synched.
- New `@since` tags are marked with `????` placeholder, to be replaced when release version is known.

## Other notes

- New JS uses `wp.utils.WordCounter()`, but the `word-count` core script is only enqueued when needed. This was throwing an error on pages where main CMB2 JS is included, but there's no fields (e.g. admin dashboard). So I include it with: `( wp.utils ? new wp.utils.WordCounter() : {} )`

## Screenshots

image

image


adriantoll:
@jtsternberg @tw2113 Hello. Is there an ETA for a decision on this? I could do with this for a project I'm working on, and wondering whether to roll my own or wait for this to land in core.

tw2113:
@adriantoll None I'm aware of. Feel free to pull in the changes to your own copy of the repo and try out. It'd also aid in testing and making sure nothing would break.

adriantoll:
@tw2113 OK, will try to do that in the next few working days and will report back.

adriantoll:
@tw2113 I've tried this on a local copy of a client site, both front and back end, and it works well.

@gyrus A couple of small points:

1. The character counter doesn't count spaces as characters - not sure if that's intentional or not. I'd expect it to count spaces.
2. char_max_msg isn't hidden on the front end to begin with, unlike the back end which is conditional on hitting the limit. That could create uncertainty because it's displayed when the text shouldn't be at risk of being truncated.

gyrus:
@adriantoll I can look into the spaces issue with the character. The code uses the WP core function from `/wp-admin/js/word-count.js`, which takes 3 types: `'words'`, `'characters_excluding_spaces'` or `'characters_including_spaces'` - and I am passing `'characters_including_spaces'` there. So something odd going on.

Regarding the `char_max_msg` on the front-end, thanks for spotting that, looks like I've been developing with `cmb2_enqueue_css` set to false. I'll add some bare-bones styling to the front-end CSS.

gyrus:
@adriantoll I've added `.cmb2-char-max-msg { display: none; }` to `_front.scss`. As far as I can see the idea in front-end styles in CMB2 is to keep to absolute essentials, so I've followed that - minimal functional CSS, nothing decorative.

Regarding the spaces in character counting, it looks like the counter's not immediately updating when you type a space at the end of the string - but it updates, including the space, once you type another non-space character. This is because the string is trimmed before the count is done. This seems right, to discount trailing spaces. Let me know if I've missed something though.

mike-weiner:
Any updates as to whether or not this is in the works to be pulled into CMB2?

gyrus:
@mike-weiner Apart from the minor issues in the CMB2 build process for this (documented above, I assume easily resolved by CMB2 devs), there's nothing from my point of view currently holding it back.

jtsternberg:
@gyrus great work on this and thank you for your patience. Thanks to all others for testing, it really helps the process and helps to confirm that this is a good addition! I will be merging and making some tweaks to the update, hopefully very soon, but feel free to bump this thread as a reminder if it lags again.

mike-weiner:
@gyrus This is freaking awesome!

Is there anyway to add in an option to his feature to limit a field to a certain type of character along with your character limit? Like alphanumeric only? Etc.?

If not, no worries! Thank you for all of the hard work!

jtsternberg:
@mike-weiner hmm I’d rather not turn this into a validation tool. That’s certainly needed, but a bit out of scope for something like this. A simple fix for validation is to use the html5 validation field types and attributes.

jtsternberg:
@gyrus there's some cleanup that needs to be done. I will do it, but I ask that you test the cleaned up version with all your existing testing procedures so you can let me know if anything breaks as a result. Some of the things I'll be cleaning up:
- redundancy in code
- `if:endif;` -- does not follow code style of CMB2 (vs `if{}`)
- enqueue snafus... `word-count` does not appear to be enqueued in all cases, and also since you're using underscore `_.debounce()`, we need to make sure that is also enqueued explicitly
- Maybe some others. Still reviewing.

gyrus:
@mike-weiner Agree with @jtsternberg that character type is extending the scope a bit here. But many thanks for your support, glad you like the feature.

@jtsternberg Great news! Thanks for the tidy-up, and of course I'm happy to thoroughly test the merged version. Keep us posted.

mike-weiner:
I am testing this in a dev. environment of mine and everything seems to be working great, except for one thing.

I am trying to overwrite the default the **char_max_msg**, but it looks as though it is not updating. (See screenshot below).

Screen Shot 2019-08-25 at 7 43 23 AM

jtsternberg:
Yes, I removed that in favor of the way we handle other text strings, as well as added the other text strings to be able to be overridden. See the `text` parameter described here: https://github.com/CMB2/CMB2/wiki/Field-Parameters#text

You can see the text ids documented in the change log entry: https://github.com/CMB2/CMB2/blob/develop/CHANGELOG.md

jtsternberg:
@gyrus it’s merged/updated. You can test with the development branch.

mike-weiner:
Ah! 

Thank you @jtsternberg !

gyrus:
@jtsternberg Current code in `develop` works well with our tests.

PR #1283, opened 2019-07-16
Added support for searchable columns

RubenMartins:
Added support to search columns automatically.

## Description
Added support to search columns automatically in admin tables list when the "column" attribute is true.

## Motivation and Context
There was a need to search columns generated by CMB2 automatically without custom code.

## Risk Level
minimal risk

## Testing procedure
Using the example-functions.php you can verify that the field with column attribute defined will automatically be able to sort.

## Types of changes

## Checklist:
- [ ] I'm facing an issue that there is a loop of 3 for this new filters witch I had to change the line 494 to "$join =" witch should be "$join .=" in order to allow further joins.
- [ ] Also the new function columns_search_where()  runs 3 times can you give an hand on this?

### DIFF
--- a/includes/CMB2_hookup.php
+++ b/includes/CMB2_hookup.php
@@ -143,6 +143,9 @@ class CMB2_hookup extends CMB2_Hookup_Base {
 				add_filter( "manage_edit-{$post_type}_sortable_columns", array( $this, 'columns_sortable' ) );
 				add_action( 'pre_get_posts', array( $this, 'columns_sortable_orderby' ) );
 			}
+			add_filter( 'posts_join', array( $this, 'columns_search_join' ) );
+			add_filter( 'posts_where', array( $this, 'columns_search_where' ) );
+			add_filter( 'posts_distinct', array( $this, 'columns_search_distinct' ) );
 		}
 
 		return $this;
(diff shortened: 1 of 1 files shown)


jtsternberg:
This is a very far-reaching change. It will cause search queries to become extremely slow in some cases, because of the JOIN, but also because of the number of meta rows which will potentially be queried, and finally because meta values are not indexed.
Also, this does not limit the meta keys to be searched, so will search every meta value associated with the posts.
I'm sorry, but I think this is better off as plugin/addon territory rather than in the core of CMB2.

RubenMartins:
I understand @jtsternberg I faced that concern during the development but would like your input on this. 
Cause I think this is a good feature to have.

I think there is two options to consider:
1. Restrict to only the CMB2 columns activated in postmeta table?
2. Allow to active this on parameters like you have done to disable the sort I can add a parameter to "activate_search_columns". And inform the slowness issue on the docs.

jtsternberg:
@RubenMartins I think both ideas, but unfortunately I'm still positive this is better off as an addon. It's a cool concept and no doubt would be helpful to a few, but it has too many downsides compared to the upsides it brings. If you're willing to develop it, I would be happy to list it in the readme.

PR #1295, opened 2019-08-26
Add description_cb to CMB2_Field

### LINKED ISSUE #1293: Move description from end of tag / Convert hardcoded description in all CMB2_Type render() functions into a hook
Hi,
So I'm trying to move the description from under the actual field html tag (input, select, textarea etc.) to show under the label (or in a tooltip. whatever) - and thus save some vertical space on my page. This lead me to dig a little into the code. Found out how to override the entire row callback, how to hook into almost every part of the cmb field and row. Move the label, override the label. etc.

Strangely enough I found that the description is hardcoded (not hooked) into the tag. Hardcoded into all  the different `CMB2_Type_*->render()` functions.

I'd like an option to unhook the description from the tag. and hook it anywhere else. The last part - adding the description somewhere else - I think is easy enough. The first part though would require rewriting all the different render() implementations.

Also from a design perspective it seems wrong that `Type_Select->render()` would output anything more than the atomic `'<select%s>%s</select>'`. Instead it prints out `'<select%s>%s</select>%s'`. That last trailing bit (the description) should be free and in the purview of whatever function is rendering everything else - the label, the th/td divs, the field input tag, etc.

abuyoyo:
## Description
Added `description_cb` callback to `CMB_Field` and changed `CMB2_Type_*::render()` methods to only print the atomic html tag and not print the description. Description now rendered exclusively by `description_cb` callback.

## Motivation and Context
Fixes #1293.

## Risk Level
admin-only = minimal risk

## Testing procedure

I did not run any unit-tests.
I did prints out a page with all field types, including group, and including one external field type ("switch") I've installed locally.
This works well with all built-in field types.
The switch field printed the description twice (obviously). One was the switch's hardcoded description, followed by the new description_cb callback. So I'm guessing this could be considered some sort of breaking change. YMMV

## Types of changes

- **Breaking change (fix or feature that would cause existing functionality to change)**
It's one of the above - that's for sure.

## Checklist:

- [x] My code (_maybe_) follows the code style of this project.

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -100,6 +100,7 @@ class CMB2_Field extends CMB2_Base {
 		'options_cb',
 		'text_cb',
 		'label_cb',
+		'description_cb',
 		'render_row_cb',
 		'display_cb',
 		'before_group',
@@ -926,6 +927,7 @@ class CMB2_Field extends CMB2_Base {
 
 		$types = new CMB2_Types( $this );
 		$types->render();
+		$this->peform_param_callback( 'description_cb' );
 
 		$this->peform_param_callback( 'after' );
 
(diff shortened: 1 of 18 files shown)


abuyoyo:
Do I need to add unit-tests to have this PR move forward?
It works. I have an admin page with all known types accounted for and they render as expected.

PR #1307, opened 2019-10-02
Add "taxonomy_select_hierarchical" in CMB2_Sanitize->default_sanitization()

latheva:
The new field type "taxonomy_select_hierarchical" did not save the data to the right table.

## Description
I have added a line of code into CMB2_Sanitize->default_sanitization() in order to save the taxonomy_* in to right table.

## Motivation and Context
When I tried to work on this new field type, it did not save the data into the right table, it suppose to update  the table _term_relationships instead it save the data into _postmeta

## Testing procedure
I have tested this code directly in my project. I have a list of custom_type taxonomies for a custom post_type. and with this modification it works as expected in my local, and prod environments.

NOTE: I have dowloaded the project from CMB2:develop branch.

### DIFF
--- a/includes/CMB2_Sanitize.php
+++ b/includes/CMB2_Sanitize.php
@@ -86,6 +86,7 @@ class CMB2_Sanitize {
 				$sanitized_value = $this->textarea();
 				break;
 			case 'taxonomy_select':
+			case 'taxonomy_select_hierarchical':
 			case 'taxonomy_radio':
 			case 'taxonomy_radio_inline':
 			case 'taxonomy_radio_hierarchical':


tw2113:
I'm not managing to recreate this postmeta table saving. It keeps saving to the `wp_term_relationships` table like logically expected.

Do you have other customizations in place that may be contributing to what issues you're seeing?

latheva:
I have re-tested right now and without this modification, the table `wp_term_relationships` doesn't got updated on my side. I have even tested for the ordinary taxonomy 'category' as well, it seems doesn't work.

jtsternberg:
👍 good catch, thank you!

PR #1339, opened 2020-02-18
Ensure to enqueue wp-color-picker

### LINKED ISSUE #1338: Can't switch between Visual and Text mode on classic editors
## Describe the bug
Can't switch between Visual and Text mode on classic editors on a clean install and only adding the CMB2 library

Error is caused by the call of farbtastic() function ([cmb2.js line 974](https://github.com/CMB2/CMB2/blob/develop/js/cmb2.js#L974))

## Possible Solution
On comment this line issue disappears
Is possible that farbtastic gets removed? a possible solution could be use wpColorPicker instead

rubengc:
For some reason, WordPress stops to enqueue wp-color-picker by default on post edit screens

This PR adds an extra check to enqueue wp-color-picker if hasn't enqueued yet

Fixes #1338

### DIFF
--- a/includes/CMB2_JS.php
+++ b/includes/CMB2_JS.php
@@ -90,6 +90,11 @@ class CMB2_JS {
 				self::colorpicker_frontend();
 			}
 
+			// Enqueue colorpicker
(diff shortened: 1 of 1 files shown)


jtsternberg:
🙌 thanks

PR #1340, opened 2020-02-18
Fixes PHP warnings on repeatable ColorPicker with an array as default

### LINKED ISSUE #1285: Multiples warnings on repeatable color picker
Hi there

I just found multiples PHP warnings when setting up a `colorpicker` field as `repeatable` and passing as `default` parameter an array

The fields get displayed correctly and initializes correctly with the values you setup, but this configuration causes a set of warnings all of them on class `CMB2_Type_Colorpicker`

2 errors "preg_match() expects parameter 2 to be string, array given" on CMB2_Type_Colorpicker lines 39 and 44
1 error "trim() expects parameter 1 to be string, array given" on CMB2_Type_Colorpicker line 46

Hope it helps to continue improving CMB2

Best regards

### LINKED ISSUE #1286: [Proposal] Support arrays on default and default_cb for repeatable fields
Actually, if you provide a default or default_cb as array on a repeatable field will result that this field will get initialized with the correct number of fields as array

A text field setup like:
```
	$cmb_demo->add_field( array(
		'name'         => __( 'Test', 'cmb2' ),
		 'type'          => 'text',
		'repeatable' => true,
		'default'       => array('Text 1','Text 2','Text 3'),
	) );
```

Actually results on a repeatable text field with 3 inputs (one for each array value) which is an awesome feature to have

On issue #1285 I was described that with this current ability I initiate a colorpicker field with a set of values but there is just one issue

Each field is created with each array value, what is ok, but the empty field (the last that CMB2 uses as pattern to duplicate) receives the full array as value

My proposal, if value given as default or default_cb is an array and field is repeatable, for the empty field pass an empty value instead of the array to avoid issue as described on #1285

rubengc:
This PR fixes PHP warnings when setting up a ColorPicker field as repeatable and passing as default parameter an array caused by pass this array to preg_match() and trim() functions

Basically, this doesn't causes any error, just PHP warnings that would be awesome to silence

Check issue #1285 where everything gets explained

Fixes #1285
Fixes #1286


tw2113:
I'd honestly prefer that the actual errors got addressed instead of just silencing them here.

jtsternberg:
I agree, I think there's a better way to ensure compatibility than telling it to shut up. :)

rubengc:
Which way do you prefer?

Checking if is an array, getting only the first value like this:
```
if( is_array( $meta_value ) ) {
    $meta_value = $meta_value[0]
} 
```

Or, only applying sanitization if is an string:
```
	if( gettype( $meta_value ) === 'string' ) {
		$hex_color = '(([a-fA-F0-9]){3}){1,2}$';
		if ( preg_match( '/^' . $hex_color . '/i', $meta_value ) ) {
			// Value is just 123abc, so prepend #
			$meta_value = '#' . $meta_value;
		} elseif (
			// If value doesn't match #123abc...
			! preg_match( '/^#' . $hex_color . '/i', $meta_value )
			// And value doesn't match rgba()...
			&& 0 !== strpos( trim( $meta_value ), 'rgba' )
		) {
			// Then sanitize to just #.
			$meta_value = '#';
		}
	}
```

rubengc:
Ah, I forgive to mention here that this issue only happens on the empty field generated by the repeatable field 

The empty field is the field used as pattern every time the field gets repeated

jtsternberg:
Best would be a new method for the sanitization, and send the meta value into the method, or do array_map to loop through the array to sanitize.

rubengc:
Ready!

tw2113:
Can you resolve the conflicts coming up in the PR here?

includes/types/CMB2_Type_Colorpicker.php
```diff
@@ -69,6 +57,45 @@ public function render( $args = array() ) {
...
+				&& 0 !== strpos( trim( $color ), 'rgba' )
+			) {
+				// Then sanitize to just #.
+				$color = '#';
+			}
+
+		}
+
+		return $color;
+
```
jtsternberg:
```suggestion
```

rubengc:
Tested with both fields and everything works perfectly

The unique issue I found is an "array to string conversion" error from the CMB2_Utils concat_attrs() function:
https://github.com/CMB2/CMB2/blob/develop/includes/CMB2_Utils.php#L584-L599

I added on this PR a change checking if is an array, turn it into a comma-separated list of values:
https://github.com/CMB2/CMB2/pull/1340/commits/42bcb971eadd15a554b415bdec622d8f30d68fe5

tw2113:
the PR is still showing some conflicts in CMB2_Type_Colorpicker.php file here in GitHub's UI.

includes/CMB2_Utils.php
```diff
@@ -589,6 +589,7 @@ public static function concat_attrs( $attrs, $attr_exclude = array() ) {
 		foreach ( $attrs as $attr => $val ) {
 			$excluded = in_array( $attr, (array) $attr_exclude, true );
 			$empty    = false === $val && 'value' !== $attr;
+			$val 	  = is_array( $val ) ? implode( ',', $val ) : $val;
 			if ( ! $excluded && ! $empty ) {
 				// if data attribute, use single quote wraps, else double.
```
jtsternberg, line 594: Move the value handling to inside the condition.
```suggestion
			if ( ! $excluded && ! $empty ) {
				$val = is_array( $val ) ? implode( ',', $val ) : $val;

				// if data attribute, use single quote wraps, else double.
```

includes/types/CMB2_Type_Colorpicker.php
```diff
@@ -69,6 +57,44 @@ public function render( $args = array() ) {
...
+	 *
+	 * @param string|array $color The color or array of colors to sanitize.
+	 *
+	 * @return string|array The color or array of colors, sanitized.
+	 **/
+	public static function sanitize_color( $color ) {
+
+		if ( is_array( $color ) ) {
+
+			$color = array_map( 'CMB2_Type_Colorpicker::sanitize_color', $color );
```
jtsternberg, line 73: Keep callback consistent w/ other callbacks in CMB2.
```suggestion
			$color = array_map( array( 'CMB2_Type_Colorpicker', 'sanitize_color' ), $color );
```

includes/types/CMB2_Type_Colorpicker.php
```diff
@@ -69,6 +57,44 @@ public function render( $args = array() ) {
...
 
+	/**
+	 * Sanitizes the given color, or array of colors.
+	 *
+	 * @since 2.7.1
+	 *
+	 * @param string|array $color The color or array of colors to sanitize.
+	 *
+	 * @return string|array The color or array of colors, sanitized.
+	 **/
```
jtsternberg, line 68:
```suggestion
	 */
```

rubengc:
All conflicts resolved, can this get added to CMB2?

rubengc:
could this one get merged? Changes requested have been added yet

jtsternberg:
Thank you!

Issue #1345, opened 2020-04-17
Field type: File_list button unique ID (buggish?)


pgroot91:
## Describe the bug

Not sure if its a bug? But can't seem to figure it out on how to solve the issue. In my project i have multiple file_list but some how the ID and NAME of the buttons come up empty. Warnings in the console that there are no unique IDs for the buttons.

## Screenshots
![image](https://user-images.githubusercontent.com/6934501/79598718-b4a78e00-80e4-11ea-8f45-e5e5ab7ddaf3.png)

## CMB2 Field Registration Code:

```php
add_action( 'cmb2_admin_init', 'amex_media_metabox' );
function amex_media_metabox() {

		/**
		 * Initiate the metabox
		 */
		$amex_media_box_options = array(
			'id'            => 'amex_media_metabox',
			'title'         => __( 'Media', 'cmb2' ),
			'object_types'  => array( 'amex_vehicle', ), // Post type
			'context'       => 'normal',
			'priority'      => 'default',
			'show_names'    => true, // Show field names on the left
			// 'cmb_styles' => false, // false to disable the CMB stylesheet
			// 'closed'     => true, // Keep the metabox closed by default
		);
				
		// Setup meta box
		$cmb = new_cmb2_box( $amex_media_box_options );
		
		$amex_media_tabs = array(
			'config' => $amex_media_box_options,
			'layout' => 'vertical', // Default : horizontal
			'tabs'   => array()
		);
		
		$amex_media_tabs['tabs'][] = array(
			'id'     => 'amex_images_tab',
			'title'  => __( 'Images', 'cmb2' ),
			'fields' => array(
				array(
					'id'      => '_thumbnail', // Saves to WP post thumbnail, allows the_post_thumbnail()
					'name'    => 'Featured Image',
					//'desc'    => 'Upload/Select an image.',
					'type'    => 'file',
					'options' => array(
						'url' => false,
					),
					'text' => array(
						'add_upload_file_text' => 'Add Image'
					),
				), array(
					'name' => 'Gallery',
					'desc' => '',
					'id'   => 'wiki_test_file_list3434',
					'type' => 'file_list',
					// 'preview_size' => array( 100, 100 ), // Default: array( 50, 50 )
					// 'query_args' => array( 'type' => 'image' ), // Only images attachment
					// Optional, override default text strings
					'text' => array(
						'add_upload_files_text' => 'Add Images', // default: "Add or Upload Files"
[...]

PR #1347, opened 2020-04-29
Update CMB2_Type_File_List.php

pgroot91:
Possible fix for #1345 from my knowing nothing will break with those changes.


jtsternberg:
Thanks!

PR #1358, opened 2020-06-19
Fix test for wysiwyg field with WP<5.3

navigatrum:
## Description
Use WP version instead of `current_theme_supports( 'html5', 'style' )`, to state if `Test_CMB2_Types::test_wysiwyg_field` should expect a rendered HTML string containing or not the attribute `type='text/css'`.

## Motivation and Context
`Test_CMB2_Types::test_wysiwyg_field` is currently failing for the builds with WP=3.8.

These [PHP=5.6 builds](https://travis-ci.org/github/navigatrum/CMB2/builds/694741818) and [PHP=5.4 builds](https://travis-ci.org/github/navigatrum/CMB2/builds/694792873) for all the WP versions, clearly suggest the tests are always failing with WP<5.3. I.e. `current_theme_supports( 'html5', 'style' )` is always false for the tested themes, but only with WP>=5.3 the rendered HTML string doesn't contain the attribute `type='text/css'`.

[This commit](https://github.com/CMB2/CMB2/commit/c4ac04db61485fcf3d693ece4244779da4a3a3c1#diff-148275844df7cd03afa4672b4b97ed76) forgot to use `$type_attr` for both the stylesheets. I [tested](https://travis-ci.org/github/navigatrum/CMB2/builds/694820529) a [commit](https://github.com/navigatrum/CMB2/commit/78db63a44a162a0e0e3991a78303c9367158ebb5) containing such integration, but it doesn't work.

So the problem depends on something changed in WP=5.3. 

#### Edit ####
Fixed typo.

### DIFF
--- a/tests/test-cmb-types.php
+++ b/tests/test-cmb-types.php
@@ -385,7 +385,7 @@ class Test_CMB2_Types extends Test_CMB2_Types_Base {
 		$this->assertHTMLstringsAreEqual(
 			'
 			<div id="wp-field_test_field-wrap" class="wp-core-ui wp-editor-wrap html-active">
-				<link rel=\'stylesheet\' id=\'dashicons-css\' href=\'' . includes_url( "css/dashicons$suffix.css?$version" ) . '\' media=\'all\' />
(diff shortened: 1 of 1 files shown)


jtsternberg:
Is this still valid? Why did you close it?

PR #1501, opened 2023-10-27
Avoid considering the value of default parameter as a callback in cmb2

nabeghe:
Some string values considered for the default parameter may be identified as a callback, although they aren't actually callbacks. so these values can't be set as defaults. For example, if we set the `end` string to the `default` param of a `select` field, we will face problems. a deprecation message is displayed and the `end` string is considered as a callback.

## Solution

by setting the value `false` for the `default_cb` param, you can avoid checking the `default` value as a callback.

### DIFF
--- a/includes/CMB2_Field.php
+++ b/includes/CMB2_Field.php
@@ -1626,7 +1626,8 @@ class CMB2_Field extends CMB2_Base {
 		}
 
 		// default param can be passed a callback as well.
-		if ( is_callable( $args['default'] ) ) {
+		// but by making the default_cb value false, you can avoid checking the default value as callback!
+		if ( $args['default_cb'] !== false && is_callable( $args['default'] ) ) {
 
 			$this->deprecated_param( __CLASS__ . '::__construct()', '2.2.3', self::DEPRECATED_CB_PARAM, 'default', 'default_cb' );
 


tw2113:
For my curiosity sake, can you provide an example field that is producing the deprecations and warnings, as well as what version of PHP you're running in?

nabeghe:
> For my curiosity sake, can you provide an example field that is producing the deprecations and warnings, as well as what version of PHP you're running in?

I'm using PHP 7.4

Let's create a select field:

```
$cmb->add_field( array(
	'name'       => 'Choose location',
	'desc'       => 'Description',
	'id'         => 'location',
	'type'       => 'select',
	'default'    => 'end',
	'options'    => array(
		'nowhere' => 'Nowhere',
		'end'     => 'End',
		'begin'   => 'Begin',
	),
) );
```

If, for example, the option 'end' exists in its options and we set the default value to 'end', this value is considered as a callback and isn't applied. Additionally, if WordPress debugging is enabled, a deprecated error will be displayed

tw2113:
Looks like this specific example is because there's an `end()` function in PHP

https://www.php.net/manual/en/function.end.php

@jtsternberg any thoughts on this one?

nabeghe:
> Looks like this specific example is because there's an `end()` function in PHP
> 
> https://www.php.net/manual/en/function.end.php
> 
> @jtsternberg any thoughts on this one?


Yes it's because of the `end` function in PHP & any string value that is equal to a function is considered as a callback
of course this problem isn't only for `select` fields & it exists in other fields as well. for example a `text` field

To apply such values, the `default_cb` param must be used which receives a callback & then returns a value. but I think the `default` & `default_cb` params aren't used together in a same field. so something can be done to prevent `default` from being checked as a callback by setting a value like `false` for  `default_cb` 🤔
The initial value of `default_cb` is an empty string, `false` will not be considered as a callback like an empty string so I don't think there will be any problem

jtsternberg:
First of all, I definitely understand the issue here. These callable defaults have been a problem for a long time, which is why they are deprecated. Unfortunately because CMB2 is so widely used, I'm wary of actually removing the code.

I understand the goal here as well, but the truth is, there's already a workaround, and that's to return the string from the `default_cb` callback. e.g.

```php
$cmb->add_field( array(
	'name'       => 'Choose location',
	'desc'       => 'Description',
	'id'         => 'location',
	'type'       => 'select',
	'default_cb' => function() { return 'end';},
	'options'    => array(
		'nowhere' => 'Nowhere',
		'end'     => 'End',
		'begin'   => 'Begin',
	)
) );
```

It's not pretty, but then again, neither is:

```php
'default_cb' => false,
'default' => 'end',
```

PR #1510, opened 2024-03-29
Jt/fix serialized values

jtsternberg:
## Description
Update how we store/retrieve serialized values (the `text_datetime_timestamp_timezone` field).

Also… bumps our min. php support to 7.4 🥳 

## Motivation and Context
security.

## Testing procedure

**Before this PR**

1. Create fields (preferably in every [core object type](https://github.com/CMB2/CMB2/blob/4fea2cc83c6111f46adc0d89d1001a7f24d593ac/includes/CMB2.php#L29)) with the `text_datetime_timestamp_timezone` field type
   * example:
   ```php
   $cmb->add_field( array(
      'name' => __( 'Test Date/Time Picker/Time zone Combo (serialized DateTime object)', 'cmb2' ),
      'desc' => __( 'field description (optional)', 'cmb2' ),
      'id'   => '_yourprefix_demo_datetime_timestamp_timezone',
      'type' => 'text_datetime_timestamp_timezone',
      'time_format' => 'H:i',
      'date_format' => 'Y-m-d',
      // 'repeatable' => true,
      'before_field' => function ( $args, $field ) {
         // echo '<xmp>'. __LINE__ .') $field: '. print_r( $field, true ) .'</xmp>';
         $value = get_post_meta( $field->object_id, $field->id(), 1 );
         echo '<xmp>$value: '. print_r( $value, true ) .'</xmp>';
         $_utc_value = get_post_meta( $field->object_id, $field->id(). '_utc', 1 );
         echo '<xmp>$_utc_value: '. print_r( $_utc_value, true ) .'</xmp>';
      },
      'column' => true,
   ) );
   ```
   And you can test the `options-page` type with the following setup: https://gist.github.com/jtsternberg/d6ab21ca3a7de1a77d1adfefa0b9445f

1. In each object type, udpate the value, save, see that it all works as expected.
1. Check the DB values -- you should see a double-serialized `DateTime` object
1. Try using `get_post_meta` to retrieve the value and see that it's a serialized `DateTime` object. (if you used the example above, you can see the value in the `before_field` callback).

**Next**
1. Pull down this PR
1. Review every field in each object type (refresh the page) and make sure field looks/acts correctly.
1. See that the `get_post_meta` value still gives you a serialized `DateTime` object.
1. Try storing a bad actor value to the DB: `wp post meta update 1 _yourprefix_demo_datetime_timestamp_timezone 'O:4:"Evil":2:{s:4:"hack";N;s:2:"me";R:2;}'`
1. Reload the various pages and see that the field still looks/acts correctly, except the value is blank.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
 		"source": "http://wordpress.org/support/plugin/cmb2"
 	},
 	"require": {
-		"php": ">5.2.4"
+		"php": ">7.4"
 	},
 	"suggest": {
 		"composer/installers": "~1.0"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "254a78029d63f104ce8cdf4ad3b8f90b",
+    "content-hash": "9aa709e984002f7435774464cd44a6f8",
     "packages": [],
     "packages-dev": [
         {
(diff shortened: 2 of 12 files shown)


tw2113 (approved):
All looked good to me overall JT.

PR #1542, opened 2025-07-30
updated color picker js

stalukder03:
Updated wp-color-picker-alpha.js from v2.1.3 to v3.0.4


js/wp-color-picker-alpha.js
```diff
@@ -4,493 +4,681 @@
...
+		 * @access private
+		 *
+		 * @return {object} The current alpha options.
+		 */
+		_getAlphaOptions: function () {
+			var el = this.element,
+				type = ( el.data( 'type' ) || this.options.type ),
+				color = ( el.data( 'defaultColor' ) || el.val() ),
+				options = {
+					alphaEnabled: ( el.data( 'alphaEnabled' ) || false ),
```
jtsternberg, line 454: **[HIGH] Breaking change: `data-alpha` attribute no longer recognized**

The old v2.1.3 library read alpha support from `this.element.data('alpha')`, which matched the `data-alpha="true"` attribute that CMB2 sets in `CMB2_Type_Colorpicker.php:62`.

This new v3 reads `el.data('alphaEnabled')` instead (i.e. `data-alpha-enabled` in HTML). Since CMB2 still sets `data-alpha`, this will be silently ignored and **alpha color picking will stop working**.

**Fix needed in `includes/types/CMB2_Type_Colorpicker.php`:**
Change `$args['data-alpha'] = 'true';` to `$args['data-alpha-enabled'] = 'true';`

js/wp-color-picker-alpha.js
```diff
@@ -4,493 +4,681 @@
...
+					alphaCustomWidth: 130,
+					alphaReset: false,
+					alphaColorType: 'rgb',
+					alphaColorWithSpace: false,
+					alphaSkipDebounce: ( !!el.data( 'alphaSkipDebounce' ) || false ),
+				};
+
+			if ( options.alphaEnabled ) {
+				options.alphaEnabled = ( el.is( 'input' ) && 'full' === type );
```
jtsternberg, line 463: **[HIGH] Additional gate: requires `data-type="full"` to enable alpha**

Even after fixing `data-alpha-enabled`, alpha will still be disabled because of this second check. The `type` variable comes from `el.data('type')` (line 451), and CMB2 never sets `data-type` on colorpicker inputs.

Since `type` will be `undefined`, the `'full' === type` check will fail and `alphaEnabled` gets forced to `false`.

**Fix needed in `includes/types/CMB2_Type_Colorpicker.php`:**
Add `$args['data-type'] = 'full';` alongside the `data-alpha-enabled` attribute.

jtsternberg (changes requested):
## PR Review: Updated wp-color-picker-alpha v2.1.3 → v3.0.4

### Verdict: REQUEST CHANGES

The library update itself looks good (cleaner architecture, RTL support, jQuery 4.0 compatibility), but **the v3 API is not backward-compatible with v2** and this PR is missing the required companion PHP changes. As-is, this will silently break alpha color picker functionality for all CMB2 users who have `'alpha' => true` on their colorpicker fields.

See inline comments for specifics.

### Additional issue not in this diff

`includes/CMB2_JS.php:151` still registers the script with version `'2.1.3'`:
```php
$func( 'wp-color-picker-alpha', ..., array( 'wp-color-picker' ), '2.1.3' );
```
This needs to be updated to `'3.0.4'` for proper cache-busting.

### Summary of required companion changes

In `includes/types/CMB2_Type_Colorpicker.php`, the alpha block (line ~60-63) needs to change from:
```php
$args['data-alpha'] = 'true';
```
to:
```php
$args['data-alpha-enabled'] = 'true';
$args['data-type']          = 'full';
```

And in `includes/CMB2_JS.php:151`, update `'2.1.3'` → `'3.0.4'`.

---
*This review was conducted with AI assistance and human oversight.*

PR #1552, opened 2026-01-20
Use WP Timezone String

lipemat:
Use `wp_timezone_string` to return value of time zones

    For sites using UTC offsets instead of timezones we need to fallback
    to an offset instead of using the option directly.

    - Fixes [#7]
    - @props angryaxi


lipemat:
Not sure how this ended up on upstream. :-(

PR #1559, opened 2026-06-01
Fix: sanitize field_id input and escape rel attribute in oEmbed handler

thisismyurl:
Hey folks,

Two related issues in `CMB2_Ajax::oembed_handler()` and `get_oembed()`:

1. `$_REQUEST['field_id']` is accessed without an `isset()` guard (PHP notice if absent) and without `sanitize_text_field()`/`wp_unslash()` before passing it into the embed args array.

2. In `get_oembed()`, that value is concatenated directly into an HTML `rel` attribute — `rel="' . $oembed['args']['field_id'] . '"` — without `esc_attr()`. WordPress Plugin Check flags this, and it's a reflected XSS vector for any admin with access to an oEmbed field.

Two-line fix: `isset()` + `sanitize_text_field( wp_unslash() )` at input, `esc_attr()` at the `rel` attribute on output.

(full disclosure, I used AI help with the testing and tweaks - Christopher)

### DIFF
--- a/includes/CMB2_Ajax.php
+++ b/includes/CMB2_Ajax.php
@@ -91,7 +91,7 @@ class CMB2_Ajax {
 		// Get embed code (or fallback link).
 		$html = $this->get_oembed( array(
 			'url'         => $oembed_url,
-			'object_id'   => $_REQUEST['object_id'],
+			'object_id'   => isset( $_REQUEST['object_id'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['object_id'] ) ) : '',
 			'object_type' => isset( $_REQUEST['object_type'] ) ? $_REQUEST['object_type'] : 'post',
 			'oembed_args' => $embed_args,
 			'field_id'    => isset( $_REQUEST['field_id'] ) && is_string( $_REQUEST['field_id'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['field_id'] ) ) : '',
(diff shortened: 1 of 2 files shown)


jtsternberg:
Pushed a few follow-up commits to the branch (TDD throughout — each behavior has a test that was watched fail before the fix):

**Reverted the `is_string()` guard (7e91829).** Thanks for chasing the bot feedback, but that guard addresses an error that can't actually occur. I checked against WP core:

- `sanitize_text_field()` → `_sanitize_text_fields()` returns `''` immediately for any array/object (`wp-includes/formatting.php`: `if ( is_object( $str ) || is_array( $str ) ) { return ''; }`).
- `wp_unslash()` → `stripslashes_deep()` → `map_deep()` is array-safe — an array passes through untouched.

So `sanitize_text_field( wp_unslash( array ) )` yields `''` with **no TypeError** on every supported WP (that early-return has been in core since WP 4.7; CMB2 supports 6.3+). The original `isset()` guard already collapses `?field_id[]=x` to the empty-string default. I restored the clean `isset()` + `sanitize_text_field( wp_unslash() )` pattern and added `test_oembed_handler_handles_array_field_id`, which drives the handler with an array `field_id` and asserts an empty `rel=""` with no error — green both with and without the guard, so it pins the behavior and documents why the guard isn't needed.

**Other commits:**
- `test`: assert `field_id` is escaped in the `rel` attribute (locks the `esc_attr()` fix; verified RED by reverting the escape).
- `fix`: guard + sanitize `$_REQUEST['object_id']` — it was read unguarded one line above the `field_id` fix, emitting the same "Undefined array key" warning when absent. TDD'd via a handler test.
- `refactor`: `wp_unslash()` `oembed_url` before sanitize, for consistency (behavior-preserving — the value is `esc_url()`'d immediately, and `esc_url()` already strips backslashes).
- `test`: reset the `CMB2_Ajax` singleton in `tear_down()` so the new handler tests don't leak state into later tests.

Full PHPUnit suite green and stable (217 tests, 708 assertions).

